PR #41, opened 2015-07-07
Actors use Hydra::Works services

jpstroop:


jpstroop:
@flyingzumwalt can you confirm that this is the fail you were expecting (and merge if so)?

Issue #47, opened 2015-07-07
Refactor Pending and Commented Actor Specs


jpstroop:
* [x] spec/actors/curation_concerns/generic_file_actor_spec.rb (see #52)
 * [x] spec/actors/curation_concerns/generic_work_actor_spec.rb (see #53)

Issue #35, opened 2015-07-06
Move Collections specs


jpstroop:
May wind up adding some functionality that should ultimately be push up to sufia-core.

 * [x] https://github.com/projecthydra-labs/sufia-core/blob/master/spec/models/collection_spec.rb
 * [x] https://github.com/projecthydra-labs/sufia-core/blob/master/spec/controllers/collections_controller_spec.rb
 * [ ] https://github.com/projecthydra-labs/sufia-core/blob/master/spec/features/collection_spec.rb
 * [ ] https://github.com/projecthydra-labs/sufia-core/blob/master/spec/features/edit_work_in_collection_spec.rb
 * [ ] https://github.com/projecthydra-labs/sufia-core/blob/master/spec/features/show_work_in_collection_spec.rb
 * [ ] https://github.com/projecthydra-labs/sufia-core/tree/master/spec/views/collections […]

jpstroop:
Also refactor to use FactoryGirl, etc..

PR #53, opened 2015-07-07
test coverage for handling of collection associations in GenericWorksActor#update

flyingzumwalt:
refs #47


flyingzumwalt:
There's not much test coverage for Collections right now.  We tore out all of the Collections stuff from Worthwhile and intended to pull in all the Collections stuff from Sufia but didn't get around to doing the latter part.

jpstroop:
Cf #35?

PR #59, opened 2015-07-08
Fixes jetty tasks so you can run them from the root of your working directory

flyingzumwalt:


jcoyne:
This will also import `solr:reindex`, which I suspect will fail if you run it. Perhaps split the jetty stuff into its own rake file?

flyingzumwalt:
The alternative is not being able to run the tests.  We should merge this and then you can submit a PR that moves the reindex task.

jcoyne:
It's easier to do it right the first time than to have to find and then go back and fix bugs.

flyingzumwalt:
Closing the PR and replacing it with a ticket #69.  Keep the branch for reference (for now).

jpstroop:
:+1:

PR #62, opened 2015-07-09
refactor audit job

eliotjordan:
Let me know if I'm going in the correct direction with this. Using [generic_file_actor] (https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/actors/curation_concerns/generic_file_actor.rb#L60) as an example, I refactored the send mail on failure functionality out of audit_job.rb. 

What I understand is that this would be implemented somewhere upstream (sufia-core?). Something like the following block would be inserted into config/initializers/sufia-core_events.rb or config/initializers/curation_concern_events.rb:

```
CurationConcerns.config.send_audituser_failure_message = lambda { |generic_file, message, subject|
  UserMessages.SendAuditUserFailureMessage(generic_file, message, subject)
}
```

And something like this would be defined elsewhere in the upstream gem:

```
module UserMessages

  # send the user a message about the failing audit
  def self.SendAuditUserFailureMessage generic_file, message, subject
    login = generic_file.depositor
    user = User.find_by_user_key(login)
    logger.warn "User '#{login}' not found" unless user
    job_user = User.audituser()
    file_title = generic_file.title.first, 
    message = "The audit run at #{log.created_at} for #{file_title} (#{uri}) failed."
    subject = 'Failing Audit Run'
    job_user.send_message(user, message, subject)
  end
end
```

### DIFF
--- a/curation_concerns-models/app/jobs/audit_job.rb
+++ b/curation_concerns-models/app/jobs/audit_job.rb
@@ -3,9 +3,6 @@ class AuditJob < ActiveFedoraIdBasedJob
     :audit
   end
 
-  PASS = 'Passing Audit Run'
-  FAIL = 'Failing Audit Run'
-
   attr_accessor :uri, :id, :path
 
   # URI of the resource to audit.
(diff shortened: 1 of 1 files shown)


curation_concerns-models/app/jobs/audit_job.rb
```diff
@@ -24,15 +21,11 @@ def run
...
-      # send the user a message about the failing audit
-      login = generic_file.depositor
-      user = User.find_by_user_key(login)
-      logger.warn "User '#{login}' not found" unless user
-      job_user = User.audituser()
-      file_title = generic_file.title.first
-      message = "The audit run at #{log.created_at} for #{file_title} (#{uri}) failed."
-      subject = FAIL
-      job_user.send_message(user, message, subject)
+      if CurationConcerns.config.respond_to?(:send_audituser_failure_message)
```
flyingzumwalt: Why `send_audituser_failure_message` instead of `after_audit_failure` or `audit_failure_action`?

eliotjordan: Yeah, that's a better name for an event.

flyingzumwalt:
Other than renaming the callback in CurationConcerns.config, this looks right to me.  Remember to move that code into sufia-core along with any test coverage that was in curation_concerns.

flyingzumwalt:
@eliotjordan did you move this code into sufia-core, or did you just record it in tickets?

eliotjordan:
@flyingzumwalt Just recorded it in ticket.

PR #64, opened 2015-07-09
fixes #63 Use GenericFile instead of CurationConcerns::GenericFile in runtime

### LINKED ISSUE #54: Adding a file creates an embargo and lease that are not requested
When I run the [add_file feature test](https://github.com/projecthydra-labs/curation_concerns/blob/master/spec/features/add_file_spec.rb), it creates a Work and GenericFile as expected.  But it also creates a lease and an embargo for both of them.  The Work lease/embargo are blank, but the GenericFile lease/embargo are populated.

### LINKED ISSUE #63: Use GenericFile model, not CurationConcerns::GenericFile, in runtime
CurationConcerns is currently dealing with GenericFiles in a way that prevents you from overriding or extending the GenericFile model.  Worthwhile was forced to [do it that way](https://github.com/projecthydra-labs/worthwhile/blob/master/worthwhile-models/app/models/concerns/curation_concern/with_generic_files.rb#L7) because of a quirk in the sufia implementation of GenericFile.  Now that we have control over the whole implementation of GenericFile, I'm fixing that, but it's going to kick up a bunch of changes around routing that could be time consuming to fix. 

# What it's doing

It's assuming that you will use CurationConcerns::GenericFile objects in work.generic_files.  It should, instead assume that you will use GenericFile objects.  

Until now, sufia-models defined a GenericFile model in the root namespace.  This made things unpredictable when other gems or the host app tried to override/extend that model's behavior.  Now in curation_concerns the GenericFile model is created by the `curation_concerns:models:install` generator. It includes `CurationConcerns::GenericFileBehavior` and allows the host application free reign to override/extend.  This means we can safely make the curation_concerns code assume that GenericFile objects will have the CurationConcerns::GenericFileBehavior even though curation_concerns does not define GenericFile.

flyingzumwalt:
Also fixes #54 (I think?)

### DIFF
--- a/app/controllers/concerns/curation_concerns/files_controller.rb
+++ b/app/controllers/concerns/curation_concerns/files_controller.rb
@@ -5,7 +5,7 @@ module CurationConcerns
     included do
       include CurationConcerns::ThemedLayoutController
       with_themed_layout '1_column'
-      load_and_authorize_resource class: CurationConcerns::GenericFile
+      load_and_authorize_resource class: ::GenericFile
       helper_method :curation_concern
       include CurationConcerns::ParentContainer
     end
--- a/app/helpers/curation_concerns/url_helper.rb
+++ b/app/helpers/curation_concerns/url_helper.rb
@@ -4,9 +4,9 @@ module CurationConcerns
     # override Blacklight so we can use our 'curation_concern' namespace
     # We may also pass in a ActiveFedora document instead of a SolrDocument
     def url_for_document doc, options = {}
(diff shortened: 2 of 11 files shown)


jpstroop:
@flyingzumwalt I think you said that you expected this to surface some routing problems. Are the not  tested or :grey_question:

flyingzumwalt:
Last time I tried to make this switch a ton of tests broke on routing
errors and it wasn't clear why. They didn't break this time. I assume it's
because I cleared up all the other stuff underneath.

PR #71, opened 2015-07-10
fix import url job specs

eliotjordan:
Fixes import url job spec and refactors user messaging out of import url job. Please take a look at the virus check section (especially line [49](https://github.com/projecthydra-labs/curation_concerns/blob/cb60438ca55b52bbdcde922b04dd9662e534f7db/spec/jobs/import_url_job_spec.rb#L49)). I think what is happening is that build_original_file clobbers the existing file (from the previous test) so that you start fresh when the job is run again. There's probably a more sophisticated way of doing this, but it passes as it is.

### DIFF
--- a/spec/jobs/import_url_job_spec.rb
+++ b/spec/jobs/import_url_job_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe ImportUrlJob do
   let(:user) { FactoryGirl.find_or_create(:jill) }
 
-  let(:file_path) { '/world.png' }
+  let(:file_path) {  fixture_path + '/world.png' }
   let(:file_hash)  {'/673467823498723948237462429793840923582'}
 
   let(:generic_file) do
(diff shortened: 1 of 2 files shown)


kevinreiss:
:+1:

spec/jobs/import_url_job_spec.rb
```diff
@@ -18,50 +18,41 @@
...
+    end
+  
+    it "should create a content datastream" do
+      expect_any_instance_of(Net::HTTP).to receive(:request_get).with(file_hash).and_yield(mock_response)
+      job.run
+      expect(generic_file.reload.original_file.size).to eq 4218
+    end
+  end
+  
+  context "when the file has a virus" do
```
flyingzumwalt: The way this is stubbed out, is it actually testing anything about the ImportUrl service?  Looks like it's just testing the virus detection behavior of generic_file, which is already tested elsewhere.

flyingzumwalt:
@kevinreiss the way `generic_file` is initialized looks right to me.  In some cases I put it in a `before(:all)` block to avoid rebuilding the objects multiple times for multiple tests, but that should be the exception not the rule.

spec/jobs/import_url_job_spec.rb
eliotjordan: I think you're right. Is this even necessary?

flyingzumwalt: Unless there is code in this Job that does something distinctive with virus
checking, you should remove it.

eliotjordan:
Removed redundant virus checking and rebased.

PR #72, opened 2015-07-10
remove mailboxer from user model

eliotjordan:
Removes mailboxer user messaging related code from user.rb. See #62 and projecthydra-labs/sufia-core/issues/6.


kevinreiss:
:+1:

PR #89, opened 2015-07-15
fixes signature for import url messaging

eliotjordan:
User messaging callback requires user to be passed in as well as generic_file. The callback code lives in sufia-core, not in curation_concerns. Curation_concerns triggers the callback but doesn't doesn't implement it. The code can be found here: projecthydra-labs/sufia-core#39.


flyingzumwalt:
@jcoyne @mjgiarlo comments before I merge this?  (same goes for #88)

mjgiarlo:
:+1:

PR #90, opened 2015-07-15
Please Comment: Renames CurationConcerns::File namespace to CurationConcerns::GenericFile

flyingzumwalt:
## What this Changes

* Renames `CurationConcerns::File` namespace to `CurationConcerns::GenericFile` namespace 
* Renames `CurationConcerns::FilesController` to `CurationConcerns::GenericFilesControllerBehavior`


jcoyne:
Can you explain why you're doing this?

flyingzumwalt:
Two reasons
 1. These modules are for mixing into GenericFileBehavior, so putting them into the GenericFile namespace makes sence
 1. It's confusing to have this in a namespace that's not GenericFile
 1. It's especially confusing for curation_concerns to have a CurationConcerns::File namespace that corresponds to Works::GenericFile objects wich are PCDM::Objects _not_ PCDM::File objects

mjgiarlo:
@flyingzumwalt So we'd be declaring modules effectively "within" a class?  That seems off to me for reasons I can't fully articulate at the moment.  Is the "Giant Caveat" listed here a concern?  http://stackoverflow.com/a/10934517  FWIW, we do see folks overriding engine-provided classes in their own apps, so I wonder if this might trip us up.

flyingzumwalt:
I don't know that we need the CurationConcerns::GenericFile class.  The
pattern in lower gems would make that CurationConcerns::GenericFile::Base
and then GenericFile could be a module.

mjgiarlo:
@flyingzumwalt OK, then :+1:

mjgiarlo:
@jcoyne Are you cool with this?

jcoyne:
@mjgiarlo +0

flyingzumwalt:
... and should I get rid of the CurationConcerns::GenericFile class so that
I can make it a module rather than a class?  I suspect that I should.

On Wed, Jul 15, 2015 at 1:50 PM, Michael J. Giarlo <notifications@github.com

mjgiarlo:
@jcoyne Can you say what your concerns are?

mjgiarlo:
@flyingzumwalt That seems like it ought to happen to me.

jcoyne:
My only concerns is that this makes more work for us upstream. I'm not seeing the benefit of making CurationConcern::GenericFile a class. But these aren't strong concerns and shouldn't necessarily hold this up.

mjgiarlo:
I believe `CurationConcerns::GenericFile` is already a class. The proposal is to make it a module to bring alignment with namespacing in other gems, IIUC.

jcoyne:
Ah, I see.

flyingzumwalt:
How about now?  Obviously needs to be squashed, but other than that, what do you think?

mjgiarlo:
:+1: once squashed, rebased, and Travised.

flyingzumwalt:
Rebased against the latest master branch.  Please do wait for travis to re-run before merging.

PR #91, opened 2015-07-15
Update hydra-head so as not to create unnecessary leases/embargos

flyingzumwalt:


jcoyne:
Is this a necessary change? We're saying "use the latest", so a bundle update should have done the same.

flyingzumwalt:
Ok.  Killing the PR & the build.

PR #94, opened 2015-07-15
repairs audit service to be compatible with hydra-works

flyingzumwalt:
@jcoyne and @cam156 you wrote this service in Sufia.  It's strange.  I did my best. Please review this.

### DIFF
--- a/curation_concerns-models/app/jobs/audit_job.rb
+++ b/curation_concerns-models/app/jobs/audit_job.rb
@@ -3,16 +3,19 @@ class AuditJob < ActiveFedoraIdBasedJob
     :audit
   end
 
-  attr_accessor :uri, :id, :path
+  attr_accessor :uri, :id, :association_name
 
   # URI of the resource to audit.
   # This URI could include the actual resource (e.g. content) and the version to audit:
   #     http://localhost:8983/fedora/rest/test/a/b/c/abcxyz/content/fcr:versions/version1
   # but it could also just be:
   #     http://localhost:8983/fedora/rest/test/a/b/c/abcxyz/content
-  def initialize(id, path, uri)
+  # @param [String] id of the parent object
+  # @param [String] association_name used to find the file within its parent object (usually "original_file")
+  # @param [String] uri of the specific file/version to be audited
+  def initialize(id, association_name, uri)
     super(id)
-    self.path = path
+    self.association_name = association_name
     self.uri = uri
   end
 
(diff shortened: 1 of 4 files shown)


mjgiarlo:
Looks like a faithful adaptation to me. :+1:

flyingzumwalt:
@mjgiarlo It was stranger (without documentation) before
* `audit_status` sometimes returns messages but sometimes returns 1 or 0, with incomplete test coverage
*  it was recording version labels instead of version uris in AuditLog. Not sure if this was intentional or a holdover from Fedora 3
* it passed around a bunch of variables like `label` but then all the code I could see just wanted that value from the file you're already passing around
... etc ...

jcoyne:
@flyingzumwalt yes, it was very inconsistent. Thanks for cleaning it up.

curation_concerns-models/app/jobs/audit_job.rb
```diff
@@ -41,12 +44,12 @@ def run_audit
...
       if fixity_ok
         passing = 1
-        ChecksumAuditLog.prune_history(id, path)
+        ChecksumAuditLog.prune_history(id, association_name)
       else
         logger.warn "***AUDIT*** Audit failed for #{uri} #{error_msg}"
         passing = 0
       end
-      ChecksumAuditLog.create!(pass: passing, generic_file_id: id, version: uri, dsid: path)
+      ChecksumAuditLog.create!(pass: passing, generic_file_id: id, version: uri, dsid: association_name)
```
jcoyne: Since we're changing everything, how about getting rid of that dsid?

jcoyne: I also wonder why we don't just pass the ID of the File itself rather than the generic_file_id + association_name.  I don't think association_name is going to work well unless it's a singular association.

jcoyne:
I'm happy to merge as is, but we should make a ticket for doing away with ChecksumAuditLog.dsid and potentially a ticket to refactor to use File#id instead of generic_file_id + association_name.

flyingzumwalt:
Let's file a ticket for the other cleanup.  Right now it's causing a bunch of failures in sufia-core.  I want to get sufia-core green before opening up any other refactoring tasks.

flyingzumwalt:
Created #95 so we don't forget that these should be revisited later.  Please merge.  It's currently just hung up on Coveralls not running.

mjgiarlo:
:ok_hand:

PR #99, opened 2015-07-16
restores action_meny _collection partial from worthwhile

flyingzumwalt:
needed by sufia-core.  test coverage will come when @jpstroop fully restores collections in curation_concerns


flyingzumwalt:
this needs more work. closing the PR for now.

PR #115, opened 2015-07-17
Move BasicMetadata out of GenericFile

jcoyne:
because its used by GenericWorks too.


jpstroop:
+1, that felt awkward.

grosscol:
This looks good to me.

PR #120, opened 2015-07-18
Remove collection partials provided by hydra-editor

jcoyne:
Removing them allows sufia tests to pass.

@jpstroop we should talk if this change will cause you grief.


mjgiarlo:
Waiting on @jcoyne or @jpstroop to confirm that this won't be problematic for Princeton.

jpstroop:
Unless there are some tests that didn't make it over (`spec/features/collection_spec.rb` seems to cover it, though) I don't see why this would cause a problem. After only a quick glance, it's not clear to me where these views live in hydra-editor.

Princeton's use case for collections is pretty light-weight/immature at the moment, so I'm tempted to +1 just to keep other issues from being blocked; if it does cause us problems it's not likely to be for a while (or tomorrow :wink:), and we can worry about it then.

mjgiarlo:
:ok:  Thanks, @jpstroop

PR #131, opened 2015-07-23
Improving controller tests

escowles:
* Adding show routes for embargo/lease so tests work
* Improving pending messages for characterization tests

### DIFF
--- a/spec/controllers/curation_concerns/generic_files_controller_spec.rb
+++ b/spec/controllers/curation_concerns/generic_files_controller_spec.rb
@@ -72,7 +72,8 @@ describe CurationConcerns::GenericFilesController do
       end
 
       context "when the file has a virus" do
-        xit "displays a flash error" do
+        it "displays a flash error" do
+          skip "pending hydra-works#89"
           expect(CurationConcerns::GenericFileActor).to receive(:virus_check).with(file.path).and_raise(CurationConcerns::VirusFoundError.new('A virus was found'))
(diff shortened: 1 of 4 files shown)


jcoyne:
I don't see where the tests are calling the Embargo or Lease show routes.  I don't think that controller provides a show route.

jcoyne:
I think the access controls are failing and it's stupidly redirecting to show, even though there is no show page.

escowles:
Yes, the access controls are stupidly redirecting to non-existent show, which results in a routing error.  Adding a show route (even with no action or views) makes the tests pass.  Is there a way to make the access control failure redirect to #index instead?

jcoyne:
I think we need something like this: https://github.com/projecthydra/sufia/commit/2101779801d9dd7700949bfce9fdd1faeca2ddd0

escowles:
@jcoyne I've updated the PR to override deny_access to handle the errors, instead of defining bogus routes.

jcoyne:
@escowles Thanks, much better now.

spec/controllers/curation_concerns/generic_files_controller_spec.rb
```diff
@@ -72,7 +72,8 @@
       end
 
       context "when the file has a virus" do
-        xit "displays a flash error" do
+        it "displays a flash error" do
+          skip "pending hydra-works#89"
```
jpstroop: :+1: to annotating why a spec is being skipped. This was a big headache in the initial code shredding project.

PR #135, opened 2015-07-27
CC models install generator will not generate GenericWork to app by default

### LINKED ISSUE #112: Don't include GenericWork in host app by default. Generate it in the tests.
... also (see #114) generate another work type that is not called GenericWork and test it with a feature test 

Note: This would mean that sufia:models:install will have to generate the GenericWork work type into the host app, since sufia actually relies on GenericWork.

This achieves 2 things:
 1. Ensures that the generator actually works, because the tests rely on it to test the GenericWork model
 1. Does not put a default GenericWork work type into host apps, which they will later have to remove  
 3. Ensures that the code functions with work types that aren't called `GenericWork`

## The Install Instructions should say:
 1. set up the rails app
 2.  run `rails g curation_concerns:install`
 3. generate a work type.  call it whatever you want.  for example `rails g curation_concerns:work generic_work`

## The test_app_generator should:
* `rails g curation_concerns:install`
* `rails delete curation_concerns:work generic_work` # delete previously generated files
* `rails delete curation_concerns:work catapult` # delete previously generated files
* `rails generate curation_concerns:work generic_work`
* `rails generate curation_concerns:work catapult` (see #114)

### LINKED ISSUE #114: Test the work generator with a feature test
Ensures that the code functions with work types that aren't called `GenericWork`  
(currently the code _does not function_ with work types that aren't called `GenericWork` 

Related to #112, but separate.

In the tests run:
* `rails delete curation_concerns:work catapult` # delete previously generated files
* `rails generate curation_concerns:work catapult`

Then run a feature test that clicks through 

* creating a Catapult object
* updating it
* adding a file to it
* adding it to a collection
* removing it from a collection
* removing the file from the Catapult object
* deleting the Catapult object

### LINKED ISSUE #125: Should CurationConcerns' Actors become concerns?
(From a conversation I had with @tampakis late yesterday) #112 has us removing GenericWork (and #124 GenericFile, potentially) from the `curation_concerns:install` generator, but we'll still have hard-coded [Actor Classes](https://github.com/projecthydra-labs/curation_concerns/tree/master/curation_concerns-models/app/actors/curation_concerns) with those names (e.g. CurationConcerns::GenericWorkActor).

Should these actors become concerns (e.g. `app/actors/concerns/generic_work_actor_behavior.rb`) that the generator mixes into a generated Actor class? For example, if the user runs:

```
rails generate curation_concerns:work catapult
```

They get `app/models/actor.rb` and `app/actors/catapult_actor.rb`, which mixes in CurationConcerns::GenericWorkActorBehavior?

@tampakis, could add an explanation of the issue you were running into on #112 yesterday that sparked this conversation?

### LINKED ISSUE #134: Works can't be added to collections through the UI
Currently only collections can be added to other collections in the interface.

In order to add a GenericWork to a collection, a user clicks "Works" or "My Works," clicks "Select an action" for a specific work and chooses "Add to collection." This is supposed to bring up a modal which allows the user to choose a collection and add the work to the collection. However, the modal html doesn't actually get added to the page.

That is because [this line](https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/catalog/_action_menu_partials/_collection.html.erb#L26) is missing for the [default action partial](https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/catalog/_action_menu_partials/_default.html.erb)

tampakis:
- Closes #112: Running `rails g curation_concerns:models:install` no longer adds the GenericWork model to the main app. Refactored to remove dependencies for an explicit `GenericWork` Curation Concern model.
  - A GenericWork view, controller, and model have been removed from the CurationConcerns application code because they can be generated with `rails g curation_concerns:work GenericWork`.
  - The tests assume that the main curation concern is a GenericWork, so the test app generator now generates a GenericWork class after installing the Curation Concerns engine. This injects an actor class (which includes WorkActorBehavior), a model class (which inherits an ActiveFedora Base class and includes GenericWorkBehavior), a controller, and a view partial.
  - Blacklight/Solr needs a track_work_path to properly render the catalog views. I removed the hard-coded `track_generic_work_path` method and metaprogrammed a [track_path method](https://github.com/projecthydra-labs/curation_concerns/compare/work_generator?expand=1#diff-0a9b4c23c6216f2b506877a8d0b3b169R27) to create a url helper for each registered curation concern.
  - I came across and fixed a bug that involved the rails application creating two works when attaching a file while initially creating a work. It was due to the `generic_file_actor` creating a second GenericWork, due the changes to the [create method in this commit](https://github.com/projecthydra-labs/curation_concerns/pull/122/files#diff-17433acc42160bb6d2751a8b39c470a5L9). Changing the order of the work_actor to attach files before creating the work was causing a GenericWork to be created in two places, once through the `super` call in the create method and in the [create_metadata method](https://github.com/projecthydra-labs/curation_concerns/compare/work_generator?expand=1#diff-4ec39f7047a04e43a5ec76c347f66246L40) of the generic_file_actor. I removed that code from the generic_file_actor, which allows a work actor of any type to only [create the work](https://github.com/projecthydra-labs/curation_concerns/compare/work_generator?expand=1#diff-17433acc42160bb6d2751a8b39c470a5L10) after the files are attached.

- Closes #114: The [work_generator_spec](https://github.com/projecthydra-labs/curation_concerns/compare/work_generator?expand=1#diff-c9aea9d988868fbc02cdd2994c4109a9R27) runs `rails g curation_concerns:work catapult`, loads the newly generated files and tests that a user can successfully create a curation concern that isn't GenericWork and add it to a collection.

- Closes #125: The actor class `GenericWorkActor` was changed to `WorkActorBehavior`, which gets mixed in to the actor so that it is consistent with the the way a generated curation concern mixes in `GenericWorkBehavior`.

- Closes #134: The default [action_menu_partial](https://github.com/projecthydra-labs/curation_concerns/compare/work_generator?expand=1#diff-a73b3248c1aafaea062607d8248912ceR26) now includes the modal html partial as it does in the [collection partial](https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/catalog/_action_menu_partials/_collection.html.erb#L26). I extended the [create_work feature spec](https://github.com/projecthydra-labs/curation_concerns/compare/work_generator?expand=1#diff-97c9d8786f5638fd05970e327a539efbR35) so that it ensures that the user can add works to collections through the UI.


jpstroop:
I believe I'm :+1: to this, but I think other :eyes: would be a good idea. @terrellt? @escowles? @mjgiarlo?

mjgiarlo:
Will take a gander after lunch!  It's number 4 in my queue.

terrellt:
I'm :+1:. Good work! I'll let @mjgiarlo take a look too, since he commented.

mjgiarlo:
Will do, soon!

mjgiarlo:
:+1: Excellent.

Issue #46, opened 2015-07-07
Refactor Pending and Commented Service Specs


jpstroop:
* [x] spec/services/create_derivatives_service_spec.rb:  _(depends on projecthydra-labs/hydra-works#132)_
 * [x] spec/services/characterization_service_spec.rb: Fixed in #51

kevinreiss:
Most of the functionality tested in create_derivatives_service_spec should be moved and tested within Hydra::Works in https://github.com/projecthydra-labs/hydra-works/issues/152.

jpstroop:
@kevinreiss @tampakis (or anyone else) projecthydra-labs/hydra-works#132 is closed, not sure of the status of  projecthydra-labs/hydra-works#152. Can this proceed? It's the only real issue left on v0.1 milestone.

kevinreiss:
Okay @jpstroop. I'll pick this back this afternoon.

kevinreiss:
@jpstroop @tampakis @mjgiarlo should both sets of tests (thumbnail and a/v transcoding) in the existing create_derivatives_service_spec.rb be migrated to hydra::works?

mjgiarlo:
@kevinreiss I'm inclined to say they should if that's where the code they test lives.

jpstroop:
Those do sound like things Hydra::Works should be doing for us, and we wouldn't want to test Hydra::Works here, so I think so (without looking closely).

tampakis:
We can remove the create_derivative_spec once https://github.com/projecthydra-labs/hydra-works/pull/183 is merged.

mjgiarlo:
https://github.com/projecthydra-labs/hydra-works/pull/183 has been merged. This is now unblocked.

mjgiarlo:
Assigning to @kevinreiss.  (Thanks, @tampakis!)

kevinreiss:
[...]

PR #150, opened 2015-08-12
Provide test coverage for a/v specs and remove CurationConcerns::CreateDerivativesService

kevinreiss:
To close out #46. 

Adds some default a/v processing to GenericFile. The RDF URIs attached to the derivatives are not ideal but are a quick solution. Two specs still marked as pending for future refactoring since Hydra::Derivatives doesn't currently handle the behavior tested in the specs when the original audio/video is itself a derivative file type.


kevinreiss:
@mjgiarlo I think if this passes muster we can commit it and close out the final issue blocking the .1 release for this gem.

curation_concerns-models/app/models/concerns/curation_concerns/generic_file/derivatives.rb
```diff
@@ -0,0 +1,33 @@
...
+  module GenericFile
+    module Derivatives
+      extend ActiveSupport::Concern
+
+      included do
+        include Hydra::Derivatives
+        # This was taken directly from Sufia's GenericFile::Derivative.
+        # Using File and ServiceFile so this class will not generate two derivatives 
+        # with the same RDF URI. Is there a mime type ontology we should be using instead?
+        directly_contains_one :ogg, through: :files, type: ::RDF::URI("http://pcdm.org/use#ServiceFile"), class_name: "Hydra::PCDM::File"
```
mjgiarlo: Where are these four lines from?  I'm not sure we should be injecting these into every `GenericFile` unless some config value is set (that's the current approach in Sufia) that lets folks express that they care about A/V transcoding. I'm all for testing that this works but I'd prefer to take an opt-in approach rather than creating four possibly extraneous DirectContainers for every GF.

spec/jobs/characterize_job_spec.rb
```diff
@@ -12,9 +12,10 @@
 
   subject { CharacterizeJob.new(generic_file.id) }
 
-  it 'spawns a CreateDerivatives job' do
+  # Now that CreateDerivativesJob calls generic_file.create_derivatives directly
+  # this test needs to be travis exempt. 
+  it 'spawns a CreateDerivatives job', unless: $in_travis do
```
mjgiarlo: The description of this test doesn't seem to match the test.  The test is ensuring that when a CharacterizeJob is run, the CharacterizationService is run.  I don't see any code that says, "when the CharacterizeJob is run, a CreativeDerivativesJob is run".

spec/models/generic_file_spec.rb
```diff
@@ -13,7 +13,7 @@
   end
 
   it "is a Hydra::Works GenericFile" do
-    expect(Hydra::Works.generic_file?(subject)).to be_truthy
+    expect(subject.works_generic_file?).to be_truthy
```
mjgiarlo: You should also be able to use a more specific matcher here:

```ruby
expect(subject).to be_works_generic_file
```

mjgiarlo:
Just a few comments.  Thanks, @kevinreiss, we're not far off now.

spec/jobs/characterize_job_spec.rb
kevinreiss: Okay. Refactored this a bit to check for this.

curation_concerns-models/app/models/concerns/curation_concerns/generic_file/derivatives.rb
kevinreiss: Sorry, my comments should have been spaced out better, didn't take those four lines from sufia. They are just following the pattern used in Hydra::Works to declare derivatives. Since where this should ultimately lives is still in flux @tampakis and I discussed it and made it part of in the test app installer which mixes it into GenericFile to keep the test coverage active in this gem.

kevinreiss: If this approach seems on the right path maybe we should make this part of the standard install generator as a flag someone can select to scaffold out a/v derivatives if they want.

Issue #146, opened 2015-08-03
Pull Single Use Links down from sufia-core


jpstroop:
Per the [code shredding halftime review](https://docs.google.com/document/d/1XWW5u2qflcpqAZDf0dtDYBD22Qqkd7L5rEQuisHvgX4/edit#heading=h.9yg4m11beddd).

___TODO:___ Reciprocal issue in sufia-core: projecthydra-labs/sufia-core#___

mjgiarlo:
Review this gem which extracts single-use links out of Sufia:

https://github.com/projecthydra-labs/single_use_links

It might be nice to have this as a separate gem and the work has already been started here.

cam156:
@mjgiarlo  That gem seems to be a stub for a future gem (there is no real code).  Seems like that is above and beyond this ticket.

mjgiarlo:
Oh, hey, whaddya know. I thought @cbeer extracted single-use links at some point, but that's not it. Maybe it's all a figment of my imagination! :pill:  Thanks for reviewing that gem, @cam156

jcoyne:
@cam156 can this close now?

grosscol:
Closed by #151

PR #151, opened 2015-08-14
Adding in the single use link model, generation controller, and view controller

cam156:
see #146 

This code is copied from Sufia.  The controllers, models, and views for generating and showing sungle use links are in this PR.  The javascript used in sufia to generate the link in the dashboard and copy it to the user clipboard is not.

### DIFF
--- a/app/controllers/concerns/curation_concerns/single_use_links_viewer_controller_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/single_use_links_viewer_controller_behavior.rb
@@ -1,0 +1,70 @@
+module CurationConcerns
+  module SingleUseLinksViewerControllerBehavior
+    extend ActiveSupport::Concern
+    include CurationConcerns::DownloadBehavior
+    include Blacklight::Base
+    include Hydra::Controller::SearchBuilder
+
+    included do
+      include ActionDispatch::Routing::PolymorphicRoutes
+
+      skip_before_action :authorize_download!, only: :show
+      rescue_from CurationConcerns::SingleUseError, with: :render_single_use_error
+      rescue_from CanCan::AccessDenied, with: :render_single_use_error
+      rescue_from ActiveRecord::RecordNotFound, with: :render_single_use_error
+      class_attribute :presenter_class
+      self.presenter_class = CurationConcerns::GenericFilePresenter
+      copy_blacklight_config_from(::CatalogController)
+    end
+
+    def download
+      raise not_found_exception unless single_use_link.path == main_app.download_path(id: @asset)
+      send_content
+    end
+
+    def show
+      _, document_list = search_results({ id: single_use_link.itemId }, [:find_one])
+      curation_concern = document_list.first
+
+      # Authorize using SingleUseLinksViewerController::Ability
+      authorize! :read, curation_concern
+
+      raise not_found_exception unless single_use_link.path == polymorphic_path([main_app, :curation_concerns, curation_concern])
+
+      # show the file
+      @presenter = presenter_class.new(curation_concern, current_ability)
+
+      # create a dowload link that is single use for the user since we do not just want to show metadata we want to access it too
+      @su = single_use_link.create_for_path main_app.download_path(curation_concern.id)
+      @download_link = curation_concerns.download_single_use_link_path(@su.downloadKey)
+    end
+
+    protected
+
+      # This is called in a before filter. It causes @asset to be set.
+      def authorize_download!
+        authorize! :read, asset
+      end
+
+      def single_use_link
+        @single_use_link ||= SingleUseLink.find_by_downloadKey!(params[:id])
+      end
+
+      def not_found_exception
+        CurationConcerns::SingleUseError.new('Single-Use Link Not Found')
+      end
+
+      def asset
+        @asset ||= ActiveFedora::Base.find(single_use_link.itemId)
+      end
+
+      def current_ability
+        @current_ability ||= SingleUseLinksViewerController::Ability.new current_user, single_use_link
+      end
+
+      def render_single_use_error(exception)
+        logger.error("Rendering PAGE due to exception: #{exception.inspect} - #{exception.backtrace if exception.respond_to? :backtrace}")
+        render template: '/error/single_use_error', layout: "error", formats: [:html], status: 404
+      end
+  end
+end
--- a/app/controllers/curation_concerns/single_use_links_viewer_controller.rb
+++ b/app/controllers/curation_concerns/single_use_links_viewer_controller.rb
@@ -1,0 +1,21 @@
+module CurationConcerns
+  class SingleUseLinksViewerController < ApplicationController
+    include CurationConcerns::SingleUseLinksViewerControllerBehavior
+
+    class Ability
+      include CanCan::Ability
+
+      attr_reader :single_use_link
+
+      def initialize(user, single_use_link)
+        @user = user || ::User.new
+
+        @single_use_link = single_use_link
+
+        can :read, [ActiveFedora::Base, SolrDocument] do |obj|
+          single_use_link.valid? && single_use_link.itemId == obj.id && single_use_link.destroy!
+        end if single_use_link
+      end
+    end
+  end
+end
--- a/app/views/curation_concerns/single_use_links/new_download.html.erb
+++ b/app/views/curation_concerns/single_use_links/new_download.html.erb
@@ -1,0 +1,5 @@
+<div class="single-use-link">
+<h1>Single Use Link</h1>
+<p>Anyone can use the following link once to download the file</p>
+<%= link_to @asset, @link, class: 'download-link', data: { 'no-turbolink' => ''}, target: '_blank' %>
+</div>
--- a/app/views/curation_concerns/single_use_links/new_show.html.erb
+++ b/app/views/curation_concerns/single_use_links/new_show.html.erb
@@ -1,0 +1,5 @@
+<div class="single-use-link">
+<h1>Single Use Link</h1>
+<p>Anyone can use the following link once to view the file metadata</p>
+<%= link_to @asset, @link, target: '_blank' %>
+</div>
--- a/app/views/curation_concerns/single_use_links_viewer/show.html.erb
+++ b/app/views/curation_concerns/single_use_links_viewer/show.html.erb
@@ -1,0 +1,10 @@
+<div >
+  <h1 class="lower"><%= @asset %></h1>
+    <h2 class="non lower">Actions</h2>
+    <p>
+      <%= link_to "Download (can only be used once)", @download_link, target: '_blank', data: { 'no-turbolink' => ''} %>
+    </p>
+  <h2> Descriptions:</h2>
+
+  <%= render 'curation_concerns/base/attributes' %>
+</div>
--- a/app/views/layouts/error.html.erb
+++ b/app/views/layouts/error.html.erb
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html lang="<%= t("curation_concerns.document_language") %>">
+  <head>
+    <meta charset="utf-8" />
+
+    <title><%= h(application_name) %>: System Error</title>
+    <!-- application css -->
+    <%= stylesheet_link_tag 'application' %>
+
+    <!-- application js -->
+    <%= javascript_include_tag 'application' %>
+
+    <%= render partial: 'shared/ga', formats: [:html] %>
+  </head>
+
+<body>
+<div  id="wrapper">
+  <div class="container-fluid">
+    <div id="page-positioner">
+      <div id="content-wrapper">
+        <div id="content" class="row">
+            <div class="col-sm-8 col-md-8 col-lg-8">
+              <%= yield %>
+            </div>
+        </div><!-- /#content -->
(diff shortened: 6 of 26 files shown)


app/controllers/curation_concerns/single_use_links_viewer_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      attr_reader :single_use_link
+
+      def initialize(user, single_use_link)
+        @user = user || ::User.new
+
+        @single_use_link = single_use_link
+
+        can :read, ActiveFedora::Base do |obj|
+          single_use_link.valid? && single_use_link.itemId == obj.id && single_use_link.destroy!
+        end if single_use_link
```
grosscol: This looks like a rubocop-ism.

app/views/layouts/error.html.erb
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html lang="<%= t("sufia.document_language") %>">
```
grosscol: These localization references need to be curation_concerns.  Also, I couldn't find the localization string document_language.

spec/controllers/curation_concerns/single_use_links_viewer_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'spec_helper'
+
+describe CurationConcerns::SingleUseLinksViewerController do
+  routes { CurationConcerns::Engine.routes }
+  let(:file) do
+    GenericFile.create do |file|
+      file.add_file(File.open(fixture_path + '/world.png'), path: 'content', original_name: 'world.png', mime_type: 'image/png')
```
grosscol: Should this be Hydra::Works::AddFileToGenericFile?  I couldn't find where add_file gets define or comes from.

app/controllers/curation_concerns/single_use_links_viewer_controller.rb
cam156: @grosscol I do not think it is a mistake.  It means only run the block if single_use_link is defined.

app/views/layouts/error.html.erb
cam156: @grosscol Good catch.  I'll fix these up.

spec/controllers/curation_concerns/single_use_links_viewer_controller_spec.rb
cam156: @grosscol it is defined here: https://github.com/projecthydra/hydra-head/blob/master/hydra-core/app/models/concerns/hydra/model_methods.rb#L16-L22
I'm not sure what methods exists in hydra works, so you may be correct.

app/controllers/curation_concerns/single_use_links_viewer_controller.rb
mjgiarlo: My $0.02is not to sweat the rubocopisms 'til #152 is merged (not ready yet). I'll rebase on master and take care of all offenses in one fell swoop, so no worries here.

cam156:
@grosscol Using the Hydra::Works showed me I was using the incorrect download module, so very good catch.  I believe it is all fixed up.

mjgiarlo:
I'd like one or more of the Princetonians to review this before I :shipit: 

Thanks, @cam156!

mjgiarlo:
Even future Princetonians. ;)  @terrellt @escowles @tampakis @jpstroop @kevinreiss

tampakis:
@mjgiarlo I got it.

app/views/layouts/error.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <meta charset="utf-8" />
+
+    <title><%= h(application_name) %>: System Error</title>
+    <!-- application css -->
+    <%= stylesheet_link_tag 'application' %>
+
+    <!-- application js -->
+    <%= javascript_include_tag 'application' %>
+
+    <%= render partial: '/ga', formats: [:html] %>
```
tampakis: 'shared/ga'

app/views/layouts/error.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  </head>
+
+<body>
+<div  id="wrapper">
+  <div class="container-fluid">
+    <div id="page-positioner">
+      <%= render partial: '/masthead', formats: [:html] %>
```
tampakis: Remove this line. Curation Concerns doesn't have a masthead partial.

app/views/layouts/error.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <div id="content-wrapper">
+        <div id="content" class="row">
+            <div class="col-sm-8 col-md-8 col-lg-8">
+              <%= yield %>
+            </div>
+        </div><!-- /#content -->
+      </div><!-- /#content-wrapper -->
+      <%= render partial: '/footer' %>
```
tampakis: 'shared/footer'

cam156:
@tampakis Thanks for the review.  I will fix up what you found.

app/views/curation_concerns/single_use_links_viewer/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    <h2 class="non lower">Actions</h2>
+    <p>
+      <%= link_to "Download (can only be used once)", @download_link %>
+    </p>
+  <h2> Descriptions:</h2>
+
+  <div class="well active" id="descriptions_display">
+    <table class="table table-striped">
+      <tbody>
+      <% present_terms(@presenter) do |r, term| %>
```
tampakis: Curation Concerns doesn't use the presenter to render the Generic File metadata. Lines 7 through 21 can be replaced with:
```
<%= render "curation_concerns/base/attributes", curation_concern: @asset %>
```

app/controllers/concerns/curation_concerns/single_use_links_viewer_controller_behavior.rb
```diff
@@ -0,0 +1,64 @@
...
+    include CurationConcerns::DownloadBehavior # was Sufia::Download...
+    included do
+      include ActionDispatch::Routing::PolymorphicRoutes
+
+      skip_before_action :load_file, except: :download
+      rescue_from CurationConcerns::SingleUseError, with: :render_single_use_error
+      rescue_from CanCan::AccessDenied, with: :render_single_use_error
+      rescue_from ActiveRecord::RecordNotFound, with: :render_single_use_error
+      class_attribute :presenter_class
+      self.presenter_class = CurationConcerns::GenericFilePresenter
```
tampakis: I think that the presenter-based stuff (L12, 13, 25, 34-36) can be removed since it's not needed for the the single-use show view.

app/views/curation_concerns/single_use_links/new_download.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="single-use-link">
+<h1>Single Use Link</h1>
+<p>Anyone can use the following link once to download the file</p>
+<%= link_to @asset, @link, class: 'download-link' %> <%= link_to raw('<i class="glyphicon glyphicon-link large-icon"></i>'), '#', class: 'copypaste itemicon itemcode', title: 'Copy File URL', id: "copy_link_#{@asset.id}" %>
```
tampakis: The `link_to_raw` bit refers to javascript code that isn't there so that bit should be removed. The download links only work when they are opened in a new tab so I think ` target: '_blank'` should be provided as an option after `class: 'download-link'`

app/views/curation_concerns/single_use_links_viewer/show.html.erb
```diff
@@ -0,0 +1,21 @@
+<div >
+  <h1 class="lower"><%= @asset %></h1>
+    <h2 class="non lower">Actions</h2>
+    <p>
+      <%= link_to "Download (can only be used once)", @download_link %>
```
tampakis: ` target: '_blank'` here as well to prevent the error page from showing.

tampakis:
@cam156 I added a few more comments. I'll merge when they've been addressed. This looks great. Thanks!

cam156:
@tampakis Are we sure we do not want to head down the presenter route?  Those were introduced relatively recently to sufia and people seem to like them.  I was not around for the conversations you had so I'm good using the base view, just double checking where we are heading. @mjgiarlo @jpstroop you may have some insight into presenter or no presenter.

jcoyne:
@tampakis Presenters are very useful, especially in the case that someone wants to customize the metadata. Furthermore they simplify the interface between the view and the controller.  I've got a very big :+1: for this.

mjgiarlo:
:+1: to using presenters.

tampakis:
The presenters do look cool. It's just that they aren't fully implemented in CC. It doesn't have the `present_terms` [helper method](https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/app/helpers/generic_file_helper.rb#L7) or a `PresenterRenderer` [class](https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/app/presenters/sufia/presenter_renderer.rb). I thought it would be easier to use the already adequate template partial than to refactor generic_file, work, and collection to use presenters at this point.

app/views/curation_concerns/single_use_links_viewer/show.html.erb
tampakis: This bit here at line 12 is what doesn't work because of the missing `present_terms` [helper method](https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/app/helpers/generic_file_helper.rb#L7), which calls the missing `PresenterRenderer` [class](https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/app/presenters/sufia/presenter_renderer.rb).

jcoyne:
The `present_terms` and `PresenterRenderer` are for the automatic conversion of the presenter into an HTML form. But that feature is in no way necessary to effectively use presenters.  Presenters live at the View level, so there is no need to refactor anything at the model level (unless someone put view logic there that should be moved into a more appropriate place)

tampakis:
I'm just saying that if the presenter is used here for the single-use view, it should be applied to the other views as well.

jcoyne:
@tampakis the journey of a thousand miles begins with a single step.  You can't require someone to refactor the whole app at one time in order to get a little pull request done. If that were a rule, there would be no hydra.

jcoyne:
@cam156 can you please add a test for `app/views/curation_concerns/single_use_links_viewer/show.html.erb`  It doesn't look like it'll work (calling a non-existant method)

mjgiarlo:
Perhaps we can resolve the difference of opinions about presenters by 1) starting with *this* presenter in the PR, and 2) ticketing the creation of other presenters so that CC is consistent? cc: @jcoyne @tampakis

tampakis:
:+1:

mjgiarlo:
:clap: 

@tampakis That @jpstroop guy hinted that you might be inclined to put in those tickets. True?

app/views/curation_concerns/single_use_links/new_show.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="single-use-link">
+<h1>Single Use Link</h1>
+<p>Anyone can use the following link once to view the file metadata</p>
+<%= link_to @asset, @link %>
```
tampakis: @cam156 for some reason this link also gives an error without including `target: '_blank'` when using the presenter class.

cam156:
@tampakis @jcoyne @mjgiarlo and others...

Added tests for the views, which showed the error @jcoyne was expecting.  Added in a minimum set of fields for the presenter and got all the test running.

jcoyne:
Can we please merge #200 before this one? These two PRs are going to clash real hard and I think it'll be easier to rebase #151 on top of #200 instead of vice versa.  I'll volunteer to do the rebase if @cam156 doesn't want to.

cam156:
@jcoyne This one has already been rebased 3 times.  Thats the issue with having a PR open for two weeks.  Unfortunately I will not have any time to rebase it again today, so if the other item gets merged @jcoyne go for it.

mjgiarlo:
If @jcoyne is offering to rebase this, I'll take him up on that. ;)

app/controllers/concerns/curation_concerns/single_use_links_viewer_controller_behavior.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def download
+      raise not_found_exception unless single_use_link.path == main_app.download_path(id: @asset)
+      send_content
+    end
+
+    def show
+      raise not_found_exception unless single_use_link.path == polymorphic_path([main_app, :curation_concerns, @asset])
```
jcoyne: How does `@asset` get loaded on this line?

jcoyne: Nevermind, must be download_behavior

jcoyne:
@mjgiarlo rebased.

cam156:
@jcoyne Thanks! @mjgiarlo @tampakis Good to merge?

tampakis:
Looks good to me.

PR #169, opened 2015-08-19
Use the main_app routes for new_user_session_path. Fixes #168

### LINKED ISSUE #168: Can't find new_user_session_path
```
```
ActionView::Template::Error (undefined local variable or method `new_user_session_path' for #<#<Class:0x007f7c9d8a9b90>:0x00000005469da8>):
    2:   <%= render 'shared/add_content' %>
    3:   <%= render 'shared/my_actions' %>
    4: <% else %>
    5:   <%= link_to 'Log In', new_user_session_path, class: 'btn btn-primary login', role: 'menuitem' %>
    6: <% end %>
  /opt/goldenseal/shared/bundle/ruby/2.2.0/bundler/gems/curation_concerns-b326ca12bb8a/app/views/shared/_site_actions.html.erb:5:in `__opt_goldenseal_shared_bundle_ruby_______bundler_gems_curation_concerns_b___ca__bb_a_app_views_shared__site_actions_html_erb__1420409845893598810_70086594714280'
```

jcoyne:


jpstroop:
@jcoyne is there test coverage for this? It's ringing a bell w/ me that this was giving @tampakis or @kevinreiss trouble at some point.

tampakis:
This shouldn't be a problem. There's actually a mix of some parts of the gem specs calling `main_app.new_user_session_path` and simply `new_user_session_path`. I they should all include main_app. That would mean updating the [sessions_helpers sign_in method](https://github.com/projecthydra-labs/curation_concerns/blob/master/spec/support/features/session_helpers.rb#L34), which covers this change in the feature specs, as well as the [downloads_controller_spec](https://github.com/projecthydra-labs/curation_concerns/blob/master/spec/controllers/downloads_controller_spec.rb#L32).

jcoyne:
When the gem specs call new_user_session_path they are already in the main_app context.   They don't need to be changed.  The problem is rendering the view in the the curation_concern engine context and then calling a route helper that only appears in the main_app.

PR #171, opened 2015-08-19
redis config should set the proper namespace

jcoyne:
After forking in passenger.  Otherwise the jobs will get pushed to the default (`resque`) namespace.  Same as https://github.com/projecthydra/sufia/commit/97944c038d22131b1e57a3867c2a5f3a2d859178


mjgiarlo:
:+1:

PR #180, opened 2015-08-20
Removing calls to add_file and calling HydraWorks::AddFileToGenericFile instead

cam156:


spec/models/generic_file_spec.rb
```diff
@@ -76,8 +76,10 @@
     end
 
     it "should have attached content" do
-      subject.add_file(File.open(fixture_path + '/world.png'), path: 'content', original_name: 'world.png')
-      expect(subject.content).to be_kind_of ActiveFedora::File
+      subject.apply_depositor_metadata('jcoyne')
+      subject.save
```
mjgiarlo: Why do we need lines 79-80?

mjgiarlo:
I have one question but am otherwise :+1:

spec/models/generic_file_spec.rb
cam156: The subject needs to be saved before the service will work.  To save the file I needed to apply depositor Metadata.

jcoyne: I agree with @cam156 that you have to save before you can attach a directly_contained file, but I don't understand why this test belongs in curation_concerns in the first place.

PR #184, opened 2015-08-20
Use a non-nil name for characterization. Fixes #183

### LINKED ISSUE #183: Warning because mime_type is nil before characterization
`WARN: Unable to find a registered mime type for nil on http://127.0.0.1:8080/fedora/rest/prod/0r/96/73/73/0r967373m`
Which is logged here: 
https://github.com/projecthydra/hydra-derivatives/blob/master/lib/hydra/derivatives/extract_metadata.rb#L19

In curation_concerns we're delegating mime_type to the output of characterization here:
https://github.com/projecthydra-labs/curation_concerns/blob/6cd47b5173305e37ba7563850c3deaea09b60f45/curation_concerns-models/app/models/concerns/curation_concerns/generic_file/characterization.rb#L7

it should be delegated to the original_file though.

jcoyne:


curation_concerns-models/app/services/curation_concerns/characterization_service.rb
```diff
@@ -2,9 +2,6 @@ module CurationConcerns
   # Run FITS to gather technical metadata about the content and the full text.
   # Store this extracted metadata in the characterization datastream.
   class CharacterizationService
-    include Hydra::Derivatives::ExtractMetadata
```
mjgiarlo: What are these three deletions for?

jcoyne: That provided `filename_for_characterization` which is no longer used. `mime_type` and `uri` were required by `Hydra::Derivatives::ExtractMetadata`

mjgiarlo: :ok:

PR #187, opened 2015-08-20
Revert 1f5cff4. Transcode videos. Fixes #186

### LINKED ISSUE #186: Videos are not being transcoded
So if you upload one, you get a player with no video:

screen shot 2015-08-20 at 3 10 17 pm

This was probably caused by https://github.com/projecthydra-labs/curation_concerns/commit/1f5cff4312e3c0aa297d93a77560a2462b93b400

jcoyne:


tampakis:
@jcoyne - @kevinreiss and I took the A/V derivatives out of the models because we were unsure if they should live here in curation_concerns or up at sufia-core. We used some probably bogus rdf uri types and added that behavior so we wouldn't lose coverage in the [derivatives_spec](derivatives_spec.rb), copied from sufia.

jcoyne:
@tampakis I'm not using sufia-core, but I want to have audio and video derivatives. Does that answer the question?

tampakis:
Not all users are expecting video derivatives. They can be configured by adding the directly_contains/makes_derivatives code to the generated `app/models/generic_file.rb`.

jcoyne:
@tampakis Isnt' that what this config flag is for? https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/lib/generators/curation_concerns/models/templates/config/curation_concerns.rb#L99

Furthermore the layouts are expecting the derivatives to be there: https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/curation_concerns/generic_files/_media_display.html.erb#L10-L18

So, presently curation concerns looks very broken: #186

tampakis:
@jcoyne ok. I didn't know those views were here. I'm going to merge this and later it can be determined where these derivatives/views should ultimately live.

jcoyne:
@tampakis I'd really like that to live here. They were a part of Curate and then later Worthwhile, so it makes sense to me that they should remain here as well as this gem is a direct descendant of those projects.  I have several projects based on worthwhile that I need an upgrade path, and I don't think Sufia is what is wanted.

PR #193, opened 2015-08-22
Consolidate config files

### LINKED ISSUE #156: consolidate config files
It's confusing to have both: `config/curation_concerns.rb` and `config/curation_concerns_config.rb`

jcoyne:
Instead of two config files, use just one. Fixes #156


mjgiarlo:
:+1:

PR #194, opened 2015-08-22
Store derivatives on local file system instead of in repo.

### LINKED ISSUE #188: audio/video derivatives have bogus use predicates
Two potential solutions:
1) Don't store derivatives in the repository. Store directly on disk.
2) Update `directly_contains_one` to scope by use predicate and mime-type.

jcoyne:
Fixes #188

### DIFF
--- a/app/controllers/concerns/curation_concerns/download_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/download_behavior.rb
@@ -13,8 +13,13 @@ module CurationConcerns
     # Render the 404 page if the file doesn't exist.
     # Otherwise renders the file.
     def show
-      if file
+      case file
+      when ActiveFedora::File
+        # For original files that are stored in fedora
         super
+      when String
+        # For derivatives stored on the local file system
+        send_file file, type: mime_type_for(file), disposition: 'inline'
       else
         render_404
       end
@@ -28,17 +33,18 @@ module CurationConcerns
         authorize! :read, asset
       end
 
-      # Overrides Hydra::Controller::DownloadBehavior#load_file, which is hard-coded to assume files are in BasicContainer (PCDM Objects use direct containment)
+      # Overrides Hydra::Controller::DownloadBehavior#load_file, which is hard-coded to assume files are in BasicContainer.
       # Override this method to change which file is shown.
       # Loads the file specified by the HTTP parameter `:file`.
       # If this object does not have a file by that name, return the default file
       # as returned by {#default_file}
-      # @return [ActiveFedora::File] the file
+      # @return [ActiveFedora::File, String, NilClass] Returns the file from the repository or a path to a file on the local file system, if it exists.
       def load_file
         file_reference = params[:file]
         return default_file unless file_reference
-        association = dereference_file(file_reference)
-        association.reader if association
+
+        file_path = CurationConcerns::DerivativePath.derivative_path_for_reference(asset, file_reference)
+        File.exist?(file_path) ? file_path : nil
       end
 
       def default_file
@@ -52,6 +58,10 @@ module CurationConcerns
       end
 
     private
+
+      def mime_type_for(file)
+        MIME::Types.type_for(File.extname(file)).first.content_type
+      end
 
       def dereference_file(file_reference)
         return false if file_reference.nil?
--- a/curation_concerns-models/app/services/curation_concerns/derivative_path.rb
+++ b/curation_concerns-models/app/services/curation_concerns/derivative_path.rb
@@ -1,0 +1,27 @@
+module CurationConcerns
+  class DerivativePath
+    class << self
+      # Path on file system where derivative file is stored
+      def derivative_path_for_reference(object, destination_name)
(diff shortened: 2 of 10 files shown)


app/controllers/concerns/curation_concerns/download_behavior.rb
```diff
@@ -13,10 +13,18 @@ def default_content_path
     # Render the 404 page if the file doesn't exist.
     # Otherwise renders the file.
     def show
-      if file
+      unless file
```
mjgiarlo: Since you're already doing a type check on `file` in the `case` statement, can you move this into the `case` statement?

```ruby
case file
when ActiveFedora::File
#...
when nil
#...
else
#...
end
```

curation_concerns-models/app/services/curation_concerns/derivative_path.rb
```diff
@@ -0,0 +1,23 @@
...
+  class DerivativePath
+    # Path on file system where derivative file is stored
+    def self.derivative_path_for_reference(object, destination_name)
+      destination_name = destination_name.gsub(/^original_file_/, '')
+      _derivative_path(object, _extension_for(destination_name), destination_name)
+    end
+
+    def self._derivative_path(object, extension, destination_name)
+      file_name = destination_name + extension
+      File.join(CurationConcerns.configuration.derivatives_path, object.id, file_name)
```
mjgiarlo: I suspect this should change to use `config` instead of `configuration` since the PR consolidating configs was merged.

curation_concerns-models/app/services/curation_concerns/derivative_path.rb
```diff
@@ -0,0 +1,23 @@
+module CurationConcerns
+  class DerivativePath
+    # Path on file system where derivative file is stored
+    def self.derivative_path_for_reference(object, destination_name)
+      destination_name = destination_name.gsub(/^original_file_/, '')
+      _derivative_path(object, _extension_for(destination_name), destination_name)
+    end
+
+    def self._derivative_path(object, extension, destination_name)
```
mjgiarlo: I assume the leading underscore is a way to mark this method as private for other developers? Would declaring these methods as private (whether by defining them on the singleton class of `DerivativePath` or using `private_class_method`) break anything?

spec/controllers/downloads_controller_spec.rb
```diff
@@ -44,14 +44,19 @@
 
       context "with an alternative file" do
         context "that is persisted" do
+          let(:file) { File.open(fixture_file_path('world.png'), 'rb') }
+
+          # rubocop:disable Style/Semicolon
```
mjgiarlo: Do you object to the following?

```ruby
let(:content) do
  file.rewind
  file.read
end
```

My preference is to discourage working around Rubocop unless there are reasons to do so. (If you have reasons, I'm happy to hear them.)

spec/services/persist_derivatives_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  describe '.output_file' do
+    subject { described_class.output_file(object, destination_name, &block) }
+
+    let(:object) { double(id: '123') }
+    let(:destination_name) { 'thumbnail' }
+
+    # rubocop:disable Style/Lambda
+    let(:block) { lambda { true } }
```
mjgiarlo: Does the stabby lambda not work here?

mjgiarlo:
Also, pls rebase!

spec/services/persist_derivatives_spec.rb
jcoyne: An empty stabby lambda is completely uncomprehendingly by someone who doesn't have an advanced knowledge of the language. This is an attempt to be more explicit about what we're doing.

jcoyne: Gah -- Autocorrect! Incomprehensible.

spec/controllers/downloads_controller_spec.rb
jcoyne: Why can't we use semicolons? Don't make me use && instead.

mjgiarlo: I'd prefer being explicit about this by adding 'spec/**/*' as an exclude to the Style/Semicolon cop in .rubocop.yml than littering the codebase potentially with a bunch of one-line disable/enables. That way our ruleset explicitly says, "we don't care about semicolon usage in our test suite."

spec/services/persist_derivatives_spec.rb
mjgiarlo: OK, then I'd be in favor of following the pattern I just described above: explicitly excluding the spec directory from the Style/Lambda cop in .rubocop.yml vs. using disable/enable comments. Is that :cool: ?

curation_concerns-models/app/services/curation_concerns/derivative_path.rb
jcoyne: Thanks, I didn't know about `private_class_method`

PR #197, opened 2015-08-24
Allow Rubocop to run on Ruby 2.1.

mjgiarlo:
@cam156 was seeing 18 linting errors in Rubocop with Ruby 2.1, which was preventing her from running local specs. CurationConcerns does not *explictly* state that Ruby 2.1 is required. Thus, let's make this small syntax change to let the test suite run atop Ruby 2.1.


cam156:
:+1:  Waiting on Travis

PR #204, opened 2015-08-26
Remove unused methods

jcoyne:


mjgiarlo:
`#clear_permissions!` is used in Sufia's [ContentDepositorChangeEventJob](https://github.com/projecthydra-labs/sufia-core/blob/master/app/jobs/content_depositor_change_event_job.rb#L21) which is defined here: https://github.com/projecthydra-labs/sufia-core/blob/master/sufia-models/lib/sufia/permissions/writable.rb#L33-L35

So this can go away in CC.

PR #206, opened 2015-08-26
Removed unused methods from IngestLocalFileJob

jcoyne:


mjgiarlo:
`#job_user` can also be deleted from import_url_job.rb, but :+1:

PR #209, opened 2015-08-26
Removing duplicate type filtering from Solr queries

escowles:
Fixes https://github.com/projecthydra-labs/curation_concerns/issues/201

### DIFF
--- a/app/search_builders/curation_concerns/search_builder.rb
+++ b/app/search_builders/curation_concerns/search_builder.rb
@@ -31,22 +31,4 @@ class CurationConcerns::SearchBuilder < Hydra::SearchBuilder
     return ['edit'] if blacklight_params[:works] == 'mine'
     super
   end
-
-  # This is included as part of blacklight search solr params logic
-  def filter_models(solr_parameters)
-    solr_parameters[:fq] ||= []
-    solr_parameters[:fq] << '(' + (work_clauses + collection_clauses).join(' OR ') + ')'
-  end
-
-  def work_clauses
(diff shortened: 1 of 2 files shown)


app/search_builders/curation_concerns/search_builder.rb
```diff
@@ -31,22 +31,4 @@ def discovery_permissions
...
   end
-
-  # This is included as part of blacklight search solr params logic
-  def filter_models(solr_parameters)
-    solr_parameters[:fq] ||= []
-    solr_parameters[:fq] << '(' + (work_clauses + collection_clauses).join(' OR ') + ')'
-  end
-
-  def work_clauses
-    return [] if blacklight_params.key?(:f) && Array(blacklight_params[:f][:generic_type_sim]).include?('Collection')
```
jcoyne: I think we may have wanted to keep this. to power https://github.com/projecthydra-labs/curation_concerns/blob/master/app/helpers/curation_concerns/search_paths_helper.rb#L2-L10  It's interesting that no specs are failing though.

jcoyne: Yeah, i'm almost certain we want to drop `only_curation_concerns` which was only introduced in https://github.com/projecthydra-labs/worthwhile/commit/5cb4ccd85a5d0732c3cbdeda14fd7a5614011e10 (as `only_generic_files_and_curation_concerns`)

escowles: No specs failed when either was removed (and they generate pretty similar queries), so I'm fine keeping filter_models instead.  I'll open a new PR for that.

jcoyne: Yes, that was troubling me too.

PR #210, opened 2015-08-26
Remove the GenericFile::Export module

### LINKED ISSUE #205: Remove GenericFile::Export
https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/models/concerns/curation_concerns/generic_file/export.rb

This is used by sufia, but not CurationConcerns. Should we remove it? 

/cc @mjgiarlo @cam156

jcoyne:
It's not used by CurationConcerns (it's for Sufia) and it's not tested.
Fixes #205


curation_concerns-models/lib/curation_concerns/configuration.rb
```diff
@@ -46,12 +46,6 @@ def build_identifier
       @build_identifier ||= Time.now.strftime('%Y-%m-%d %H:%M:%S')
     end
 
-    # Set some configuration defaults
```
cam156: @jcoyne why is this part of the export PR?

jcoyne: The Export module is the only place this configure option is used

cam156: :ok_hand:

PR #213, opened 2015-08-27
Move indexing to the indexer. Fixes #202

### LINKED ISSUE #202: Do we want image dimension indexed?
Originally added in: https://github.com/projecthydra-labs/curation_concerns/commit/a0cf0739db3dd32d97951bfca19b2bb645e6125a

presently living here: https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/models/concerns/curation_concerns/generic_file_behavior.rb#L35-L36

I think we can either 1) remove or 2) move to GenericFileIndexer	but it shouldn't remain where it is.

jcoyne:


cam156:
:+1:  Looks good to me once rubocop is happy.

PR #215, opened 2015-08-27
Remove unused method ImportUrlJob#job_user. Fixes #212

### LINKED ISSUE #212: #job_user can be deleted from import_url_job.rb

jcoyne:


cam156:
@jcoyne I take it there is no notifications in the curation concerns jobs.  If that is the case then :+1:

PR #219, opened 2015-08-27
Index the filename without any text transformation

jcoyne:
This enables you to do an exact find by filename without risking a
partial match as you would if you used label_tesim.

### DIFF
--- a/curation_concerns-models/app/services/curation_concerns/generic_file_indexing_service.rb
+++ b/curation_concerns-models/app/services/curation_concerns/generic_file_indexing_service.rb
@@ -3,7 +3,9 @@ module CurationConcerns
     def generate_solr_document
       super.tap do |solr_doc|
         solr_doc[Solrizer.solr_name('representative')] = object.representative
+        # Label is the actual file name. It's not editable by the user.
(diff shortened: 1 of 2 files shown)


curation_concerns-models/app/services/curation_concerns/generic_file_indexing_service.rb
```diff
@@ -3,7 +3,9 @@ class GenericFileIndexingService < ActiveFedora::IndexingService
     def generate_solr_document
       super.tap do |solr_doc|
         solr_doc[Solrizer.solr_name('representative')] = object.representative
+        # Label is the actual file name. It's not editable by the user.
         solr_doc[Solrizer.solr_name('label')] = object.label
```
cam156: @jcoyne Should this line be removed or do you want to index the field twice?

jcoyne: I assume someone wanted to be able to search for part of a filename and get a match. That is why it is indexed with `_tesim`, so I preserved that behavior.

cam156: :ok:

PR #220, opened 2015-08-27
Make 'All' tab display both Works and Collections. Story curationexperts/goldenseal#47

val99erie:


app/views/catalog/index.html.erb
```diff
@@ -10,38 +10,10 @@
...
-        <%= render_grouped_document_index %>
-      <%- else %>
-        <%= render_document_index %>
-      <%- end %>
-
-	    <%= render 'results_pagination' %>
-
-
-    <% end %>
+  <% if has_search_parameters? || @document_list.count > 0 %>
```
jcoyne: why are we testing for `@document_list.count`? Why aren't we just doing the blacklight default behavior? https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/index.html.erb#L6

jcoyne:
I think we differ on how this should work. I think that the "All" tab should display the welcome text unless you've entered a search query. I think the current behavior is that all shows the results of the default query (*:*?)

cam156:
@jcoyne ScholarSphere and Sufia has moved away from having the all tab show the welcome message.  At this point we have a separate page for welcome.  This change is more inline with that behavior.  @val99erie I'm not sure which behavior they are expecting in curation concerns though. The old behavior aligns with blacklight.

jcoyne:
@cam156 that would be a viable option if curation_concerns had a WelcomeController, but it doesn't at this point.

val99erie:
@jcoyne & @cam156 Thanks for the feedback.  It sounds like maybe I should write a WelcomeController to handle the welcome text separately, and then the CatalogController can just handle the search results.

mjgiarlo:
:+1: to that approach, @val99erie. I believe we call it a HomepageController in Sufia if you want to look over there for any useful/applicable stuff.

val99erie:
I added the WelcomeController.  Please review when you have time.  Thanks

mjgiarlo:
Ping @cam156

spec/factories/collections.rb
```diff
@@ -11,5 +11,9 @@
     after(:build) do |collection, evaluator|
       collection.apply_depositor_metadata(evaluator.user.user_key)
     end
+
+    trait :public do
```
cam156: @val99erie this seem like a strange change for this PR.  Is it meant to be a part?

cam156:
@val99erie looks good to me if the public trait on collection was meant to be included.

spec/factories/collections.rb
val99erie: Yes, I used the public trait in the setup of spec/features/catalog_search_spec.rb (line 5), to make a collection that has public visibility.  That's here:
https://github.com/projecthydra-labs/curation_concerns/commit/b94388e7ab2afaf0da2b04f496f9965431943c24#diff-8d02e9d055253b29a5729e0e09bc80ecR5

PR #221, opened 2015-08-28
Set the Content-Length header for downloads

jcoyne:
Without a Content-Length header on an audio/video file the browser will either
not show a scrubber or disable it.


cam156:
@jcoyne you will also get inconsistent results with larger file downloads. :+1:

PR #226, opened 2015-08-28
Update .gitignore

jcoyne:


.gitignore
```diff
@@ -1,14 +1,10 @@
...
 tmp
-*.bundle
-spec/internal
 jetty
-.rvmrc
-.ruby-*
-./spec/internal
-.DS_store
+/.rvmrc
+/.ruby-version
```
mjgiarlo: Can you also add .ruby-gemset here?

jcoyne: It has been added.

mjgiarlo: :+1:

PR #238, opened 2015-08-31
Update Derivative spec to look for filesystem derivatives. Ref #194

jcoyne:


cam156:
@jcoyne Do the derivatives get attached to the GenericFile in some way?  Does the file point to the tmp area or something?

jcoyne:
@cam156 no, there is no attachment. The way I'm looking at it is access derivatives don't belong in the repository because they can be regenerated and are not important to preserve.  This is similar to how an image server can provide many different access versions of an image, which are not preserved.

kevinreiss:
:+1: to merging this. After that we close out https://github.com/projecthydra-labs/curation_concerns/issues/46.

cam156:
@jcoyne I was wondering how the GF knew where the derivatives were located on disk?  How do I know that one is there and exists if I want to display a thumbnail for instance.  I have no problem with them living on disk instead of the repository.

PR #241, opened 2015-08-31
Generate CatalogController into the host app

### LINKED ISSUE #223: Generate the blacklight config rather than hardcoding it
Sufia generates the CatalogController into the host application, allowing the user to customize the displayed fields. https://github.com/projecthydra/sufia/blob/master/lib/generators/sufia/templates/catalog_controller.rb

CurationConcerns hard codes this in a module:
https://github.com/projecthydra-labs/curation_concerns/blob/master/app/controllers/concerns/curation_concerns/catalog_controller.rb

I think the generator approach gives the user more control and I propose we switch to that.

### LINKED ISSUE #224: Trim CatalogController
I don't think we're using the show part of the blacklight configuration (we render the `_attributes.html.erb` partial instead), so we should remove that which is not important:

https://github.com/projecthydra-labs/curation_concerns/blob/master/app/controllers/concerns/curation_concerns/catalog_controller.rb#L75-L90

jcoyne:
Fixes #223

Also get rid of unused show catalog configuration. Fixes #224


spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -4,7 +4,7 @@ class TestAppGenerator < Rails::Generators::Base
   source_root '../../spec/test_app_templates'
 
   def install_engine
-    generate 'curation_concerns:install'
+    generate 'curation_concerns:install', '-f'
```
cam156: @jcoyne why is the -f needed now?

jcoyne: To force overwrite the catalog_controller generated by hydra-head's installer.  Sufia does the same thing.

PR #242, opened 2015-08-31
Extract AttributeRenderer

jcoyne:
This reduces complexity and enables translations of attributes

### DIFF
--- a/app/views/curation_concerns/base/_attributes.html.erb
+++ b/app/views/curation_concerns/base/_attributes.html.erb
@@ -4,22 +4,20 @@
     <tr><th>Attribute Name</th><th>Values</th></tr>
   </thead>
   <tbody>
-  <%= @presenter.attribute_to_html(:description, 'Abstract') %>
-  <%= @presenter.attribute_to_html(:creator, 'Creator', { catalog_search_link: true }) %>
-  <%= @presenter.attribute_to_html(:contributor, 'Contributors', { catalog_search_link: true }) %>
-  <%= @presenter.attribute_to_html(:subject, "Subject", { catalog_search_link: true }) %>
-  <%= @presenter.attribute_to_html(:publisher, "Publisher") %>
-  <%#= @presenter.attribute_to_html(:bibliographic_citation, "Bibliographic Citation") %>
-  <%#= @presenter.attribute_to_html(:source, "Source") %>
-  <%= @presenter.attribute_to_html(:language, "Language") %>
+  <%= @presenter.attribute_to_html(:description, label: 'Abstract') %>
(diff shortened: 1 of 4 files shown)


mjgiarlo:
Deferring to @terrellt

terrellt:
So the one concern I have about AttributeRenderer is the same one we had with services - it's effectively a global function. Can we have an attributes object you can iterate over with some common interface?

terrellt:
Or a form object. Something that responds to an expected API

jcoyne:
@terrellt if you build it you can. As I see it this is just one step in improving the state of things. This is not the final evolution. As you can see rubocop is no longer warning about cyclomatic and perceived complexity around this feature.

terrellt:
@jcoyne Yes, this is better than what was there. If the community's guiding post is "better means merge" then I'm okay with it.

terrellt:
@jcoyne I was about to merge - what changed?

jcoyne:
@terrellt Just a rebase to the current master

app/views/curation_concerns/base/_attributes.html.erb
```diff
@@ -4,22 +4,22 @@
...
-  <%= @presenter.attribute_to_html(:publisher, "Publisher") %>
-  <%#= @presenter.attribute_to_html(:bibliographic_citation, "Bibliographic Citation") %>
-  <%#= @presenter.attribute_to_html(:source, "Source") %>
-  <%= @presenter.attribute_to_html(:language, "Language") %>
+  <%= @presenter.attribute_to_html(:description, label: 'Abstract') %>
+  <%= @presenter.attribute_to_html(:creator, catalog_search_link: true ) %>
+  <%= @presenter.attribute_to_html(:contributor, label: 'Contributors', catalog_search_link: true) %>
+  <%= @presenter.attribute_to_html(:subject, catalog_search_link: true) %>
+  <%= @presenter.attribute_to_html(:publisher) %>
+  <%#= @presenter.attribute_to_html(:bibliographic_citation) %>
```
terrellt: Is it worth leaving these commented fields in?

jcoyne: I think yes, because we need to have a conversation with whomever took them out in the first place and ask why.

terrellt: Can we make that a ticket and remove these? If we need them back git history should be plenty, yes?

jcoyne: Yes. Ticket made https://github.com/projecthydra-labs/curation_concerns/issues/250

PR #273, opened 2015-09-03
Update the access denied code so that it gets called

jcoyne:
This reverts some of the changes in https://github.com/projecthydra-labs/curation_concerns/commit/4038d4be78538ba20657d95a8e7dc069a6d21923

### DIFF
--- a/spec/controllers/curation_concerns/generic_files_controller_spec.rb
+++ b/spec/controllers/curation_concerns/generic_files_controller_spec.rb
@@ -214,32 +214,33 @@ describe CurationConcerns::GenericFilesController do
 
   context 'someone elses files' do
     let(:generic_file) do
-      generic_file = GenericFile.new.tap do |gf|
+      generic_file = GenericFile.create(read_groups: ['public']) do |gf|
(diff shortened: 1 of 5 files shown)


spec/controllers/curation_concerns/generic_files_controller_spec.rb
```diff
@@ -214,32 +214,33 @@
...
       end
     end
-    it 'does not let the user submit if they logout' do
-      get :new, parent_id: parent
-      expect(response).to fail_redirect_and_flash(main_app.new_user_session_path, 'You are not authorized to access this page.')
+
+    describe 'new' do
+      it 'does not let the user submit if they logout' do
+        get :new, parent_id: parent
+        expect(response).to fail_redirect_and_flash(main_app.new_user_session_path, 'You are not authorized to access this page.')
```
cam156: @jcoyne should this be unauthorized now too?  Seems like all the other actions moved to a 401.

jcoyne: I thought that was odd too, but the code never made the new method go to unauthorized. I wouldn't complain if you changed it.

PR #274, opened 2015-09-03
Refactor Ability so you can override admin functionality easier

jcoyne:


curation_concerns-models/app/models/concerns/curation_concerns/ability.rb
```diff
@@ -13,20 +13,25 @@ def curation_concerns_permissions
...
+    def admin?
+      user_groups.include? 'admin'
+    end
+
     # Add this to your ability_logic if you want all logged in users to be able
     # to submit content
     def everyone_can_create_curation_concerns
       return if current_user.new_record?
-      can :create, ::GenericFile
+      can :create, [::GenericFile, ::Collection]
```
terrellt: Should there be a spec for this addition?

jcoyne: That was no addition, I just combined two lines into one.

terrellt: Ah, missed it.

PR #277, opened 2015-09-03
use to_s method for generic_file when possible

### LINKED ISSUE #235: Use to_s for generic_file title in show views.
Generic_file show pages call `generic_file.title.first` which will fail if the generic_file title isn't set. Use the to_s method instead.

tampakis:
Closes #235 
Thanks to @terrellt for helping get the edit forms to use the presenters (for their members).

### DIFF
--- a/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
+++ b/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
@@ -59,6 +59,7 @@ module CurationConcerns::CurationConcernController
   end
 
   def edit
+    @form = CurationConcerns::FormPresenter.new(curation_concern, current_ability)
   end
 
   def update
--- a/app/forms/curation_concerns/forms/collection_edit_form.rb
+++ b/app/forms/curation_concerns/forms/collection_edit_form.rb
@@ -12,6 +12,29 @@ module CurationConcerns
       def required?(key)
         model_class.validators_on(key).any? { |v| v.is_a? ActiveModel::Validations::PresenceValidator }
       end
+
+      # @return [Hash] All generic files in the collection, file.to_s is the key, file.id is the value
+      def select_files
+        Hash[all_files]
+      end
+
+      private
+
+        def all_files
+          member_presenters.flat_map(&:file_presenters).map { |x| [x.to_s, x.id] }
+        end
+
+        def member_presenters
+          load_generic_work_presenters(model.member_ids)
+        end
+
+        # @param [Array] ids the list of ids to load
+        # @return [Array<GenericFilePresenter>] presenters for the generic files in order of the ids
+        def load_generic_work_presenters(ids)
+          return [] if ids.blank?
+          docs = ActiveFedora::SolrService.query("{!terms f=id}#{ids.join(',')}").map { |res| SolrDocument.new(res) }
+          ids.map { |id| GenericWorkShowPresenter.new(docs.find { |doc| doc.id == id }, nil) }
+        end
     end
   end
 end
--- a/app/views/curation_concerns/base/_form_representative_image.html.erb
+++ b/app/views/curation_concerns/base/_form_representative_image.html.erb
@@ -1,4 +1,4 @@
-<% if curation_concern.persisted? && curation_concern.generic_files.present? %>
+<% if curation_concern.persisted? && curation_concern.member_ids.present? %>
   <div class="row">
     <div class="col-md-12">
       <fieldset id="representative-image">
@@ -6,7 +6,7 @@
           Assign a Thumbnail
         </legend>
         <p>Select the file whose thumbnail should represent this <%= curation_concern.human_readable_type %>.</p>
-        <%= f.select( :representative, Hash[ curation_concern.generic_files.map {|file| [file.title.first, file.id] } ],{}, { class: 'form-control' } ) %>
+        <%= f.select( :representative, @form.files_hash,{}, { class: 'form-control' } ) %>
       </fieldset>
     </div>
   </div>
--- a/spec/forms/collection_edit_form_spec.rb
+++ b/spec/forms/collection_edit_form_spec.rb
@@ -19,4 +19,24 @@ describe CurationConcerns::Forms::CollectionEditForm do
       it { is_expected.to be true }
     end
   end
+
+  describe '#select_files' do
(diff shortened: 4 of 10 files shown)


app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -59,6 +56,7 @@ def show_presenter
   end
 
   def edit
+    presenter
```
jcoyne: We don't want to use a Solr document as the backer for the presenter on an edit. It should use an ActiveFedora object instead.  The ActiveFedora object may have more data in it than the Solr object, which needs to go on the form.

app/forms/curation_concerns/forms/collection_edit_form.rb
```diff
@@ -12,6 +12,31 @@ class CollectionEditForm
       def required?(key)
         model_class.validators_on(key).any? { |v| v.is_a? ActiveModel::Validations::PresenceValidator }
       end
+
+      # @return [Hash] All generic files in the collection, file.to_s is the key, file.id is the value
+      def select_files
```
jcoyne: What does this have to do with using to_s?  Should this be moved to a separate PR?

app/views/curation_concerns/base/_form_representative_image.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if curation_concern.persisted? && curation_concern.generic_files.present? %>
+<% if curation_concern.persisted? && curation_concern.member_ids.present? %>
```
jcoyne: What does this have to do with using to_s? New PR?

app/views/curation_concerns/base/_form_representative_image.html.erb
```diff
@@ -6,7 +6,7 @@
           Assign a Thumbnail
         </legend>
         <p>Select the file whose thumbnail should represent this <%= curation_concern.human_readable_type %>.</p>
-        <%= f.select( :representative, Hash[ curation_concern.generic_files.map {|file| [file.title.first, file.id] } ],{}, { class: 'form-control' } ) %>
+        <%= f.select( :representative, Hash[ @presenter.file_presenters.map {|file| [file.to_s, file.id] } ],{}, { class: 'form-control' } ) %>
```
jcoyne: Good idea, but switching to presenters belongs in a different pr. Can you just keep the 'file.title.first' => 'file.to_s'

spec/forms/collection_edit_form_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe CurationConcerns::Forms::CollectionEditForm do
-  let(:collection) { Collection.new }
+  let(:collection) { create(:collection) }
```
jcoyne: Do we need to persist here? This will make for a slower test.

app/forms/curation_concerns/forms/collection_edit_form.rb
```diff
@@ -12,6 +12,31 @@ class CollectionEditForm
...
+        Hash[
+          all_files
+        ]
+      end
+
+      private
+
+        def all_files
+          member_presenters.flat_map(&:file_presenters).map { |x| [x.to_s, x.id] }
```
tampakis: select_files needs to get the titles of all the files. There isn't a `to_s` method on generic_files (which returns the filename if there isn't a title assigned). The code is refactored to access the files through presenters.

jcoyne: Perhaps we should just add to_s on GenericFileBehavior?

app/views/curation_concerns/base/_form_representative_image.html.erb
tampakis: similar to the collection one partial, the `to_s` method returns an empty string when file is missing a title since it's only defined for presenters. Maybe it makes sense to also define a `to_s` method for generic_files

jcoyne:
@tampakis part of what I'm giving feedback on is the commit message doesn't seem to match the work that was actually done.  I do prefer smaller PRs, but if you have to do all the work in one commit, please update the commit message to explain what you did and why.

I do think its a very bad idea to back the edit form with a SolrDocument since the SolrDocument might only be a subset of things needed on the form.

terrellt:
Agreed, the edit form should be backed by the real object from Fedora - but the all_files doesn't need to be.

tampakis:
Thanks @jcoyne and @terrellt. I just removed collection presenter from the edit form. I do think that the generic_work and generic_file presenters should be used within the edit forms for collections and generic_works when fetching the files because it will be really slow to fetch all of the works from Fedora and then all files for each work.

All of the work in this PR is to prevent any issues in the views when generic_files don't have titles set. Along the way, it also introduces a performance optimization for the `form_representative_image` partials to use solr rather than fedora for a collection's/generic_works's members.

spec/forms/collection_edit_form_spec.rb
```diff
@@ -19,4 +19,24 @@
...
   end
+
+  describe '#select_files' do
+    context 'without any works/files attached' do
+      subject { form.select_files }
+      it { is_expected.to be_empty }
+    end
+
+    context 'with a work/file attached' do
+      let(:work) { create(:work_with_one_file) }
```
tampakis: Good catch about persisting the collection @jcoyne. The collection needs to be persisted in order to for the #select_files method but not for the other specs.

app/views/curation_concerns/base/_form_representative_image.html.erb
jcoyne: Now that you've reverted the changes to the Controller, does @presenter exist?

tampakis: That's going to fail.

tampakis: Hmm. So I think that this was the only part of the code to actually use the presenter in the edit action. @jcoyne, is it still a bad idea to include the presenter if it's only used to retrieve its generic_files?

jcoyne: I think we should call it `@form` and make it have the same interface as the presenter.  But instead of backing by a SolrDoc, give it a Fedora object

PR #283, opened 2015-09-03
Presenter should return more than 10 files

jcoyne:
Solr defaults to returining 10 rows. This commit sets that limit at 1000

### DIFF
--- a/app/forms/curation_concerns/forms/collection_edit_form.rb
+++ b/app/forms/curation_concerns/forms/collection_edit_form.rb
@@ -25,15 +25,7 @@ module CurationConcerns
         end
 
         def member_presenters
-          load_generic_work_presenters(model.member_ids)
-        end
-
-        # @param [Array] ids the list of ids to load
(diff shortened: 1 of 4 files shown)


app/forms/curation_concerns/forms/collection_edit_form.rb
```diff
@@ -25,15 +25,7 @@ def all_files
...
-          load_generic_work_presenters(model.member_ids)
-        end
-
-        # @param [Array] ids the list of ids to load
-        # @return [Array<GenericFilePresenter>] presenters for the generic files in order of the ids
-        def load_generic_work_presenters(ids)
-          return [] if ids.blank?
-          docs = ActiveFedora::SolrService.query("{!terms f=id}#{ids.join(',')}").map { |res| SolrDocument.new(res) }
-          ids.map { |id| GenericWorkShowPresenter.new(docs.find { |doc| doc.id == id }, nil) }
+          PresenterFactory.build_presenters(model.member_ids, GenericWorkShowPresenter, nil)
```
terrellt: Yay!

PR #287, opened 2015-09-03
ThumbnailPathService should look to DerivativePath for the thumbnail

jcoyne:


curation_concerns-models/app/services/curation_concerns/thumbnail_path_service.rb
```diff
@@ -10,7 +10,7 @@ def call(object)
         return unless representative
         if representative.audio?
           audio_image
-        elsif representative.thumbnail
+        elsif thumbnail?(representative)
           Rails.application.routes.url_helpers.download_path(object.representative, file: 'thumbnail')
```
terrellt: Won't doing object.representative hit Fedora?

jcoyne: No, representative is a simple text field, not a relationship

PR #294, opened 2015-09-04
allow works to be removed from collection show page

### LINKED ISSUE #148: Collection show view should have a "remove from collection" action for each of its generic files
Within Sufia in the Show view of a Collection you see details/metadata (title, description, thumbnail) about the collection and then see, below it, search results for the items in the collection.  Those items look like search results from the dashboard and have full pagination, query (search within collection), etc.  Each result in the list of collection contents also has a button/action to "remove item from collection"

Pull those views into curation_concerns.  This will add:

 * embedded search within collection
 * “remove from collection” 
 * etc.

Per the [code shredding halftime review](https://docs.google.com/document/d/1XWW5u2qflcpqAZDf0dtDYBD22Qqkd7L5rEQuisHvgX4/edit#heading=h.9yg4m11beddd).

___TODO:___ Reciprocal issues in sufia-core: projecthydra-labs/sufia-core#___

tampakis:
Closes #148.

### DIFF
--- a/app/helpers/curation_concerns/collections_helper.rb
+++ b/app/helpers/curation_concerns/collections_helper.rb
@@ -17,8 +17,9 @@ module CurationConcerns::CollectionsHelper
 
   # override hydra-collections
   def link_to_remove_from_collection(document, label = 'Remove From Collection')
-    link_to collections.collection_path(@collection.id, collection: { members: 'remove' },
-                                                        batch_document_ids: [document.id]), method: :put do
+    collection_id = @collection ? @collection.id : @presenter.id
+    link_to collections.collection_path(collection_id, collection: { members: 'remove' },
+                                                       batch_document_ids: [document.id]), method: :put do
       icon('minus-sign') + ' ' + label
     end
   end
--- a/app/views/catalog/_action_menu_partials/_default.html.erb
+++ b/app/views/catalog/_action_menu_partials/_default.html.erb
@@ -7,7 +7,7 @@
         <% end %>
       </li>
       <li>
-        <% if @collection # We're on the view page for @collection. -%>
+        <% if @collection || (@presenter and @presenter.collection?) # We're on the view page for @collection. -%>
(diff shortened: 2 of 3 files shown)


app/helpers/curation_concerns/collections_helper.rb
```diff
@@ -17,8 +17,9 @@ def link_to_select_collection(collectible, opts = {})
 
   # override hydra-collections
   def link_to_remove_from_collection(document, label = 'Remove From Collection')
-    link_to collections.collection_path(@collection.id, collection: { members: 'remove' },
-                                                        batch_document_ids: [document.id]), method: :put do
+    collection_id = @collection ? @collection.id : @presenter.id
```
cam156: @tampakis I see @presenter.id here and below I see @presenter.collection.  I'm wondering is this the same presenter?  @presenter.collection makes more sense to me.

cam156: NM.  That is collection? so that is not the id.

cam156:
@tampakis FYI our users did not want things that edited the collection on the show page of the collection. It seems easy to override, so I have no objections to having it.

app/helpers/curation_concerns/collections_helper.rb
tampakis: The show page uses the collection presenter and the edit page use the actual collection object. If there is a collection instance variable its id gets passed, otherwise the presenter's id gets passed.

tampakis:
@cam156 I agree with you. It makes sense that the works are removable from a collection only on the edit page. I was just doing what #148 says.

tampakis:
@cam156 I'll let you decide whether to merge. The current behavior has "Delete Generic Work" instead of "Remove From Collection" for each item in a collection on the show page.

mjgiarlo:
Since @cam156 sounds :ok_hand: with this, I'm :+1:

PR #297, opened 2015-09-04
Adding checking the work representative to the GenericFileActor. fixes #268

### LINKED ISSUE #268: Clear the representative for a Work if that GenericFile is deleted

cam156:
In addition I changed the controller to use the actor to destroy the file.  This allows the call backs to be performed in addition to checking the work representative.

### DIFF
--- a/app/controllers/concerns/curation_concerns/generic_files_controller_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/generic_files_controller_behavior.rb
@@ -66,7 +66,7 @@ module CurationConcerns
     end
 
     def destroy
-      @generic_file.destroy
+      actor.destroy
       redirect_to [main_app, :curation_concerns, @generic_file.generic_works.first], notice: 'The file has been deleted.'
     end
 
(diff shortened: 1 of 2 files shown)


curation_concerns-models/app/actors/curation_concerns/generic_file_actor.rb
```diff
@@ -101,6 +101,14 @@ def update_metadata(model_attributes, all_attributes)
     end
 
     def destroy
+      # check the works to make sure the representative file for a work is not the current file
+      generic_file.generic_works.each do |work|
+        if work.representative == generic_file.id
+          work.representative.nil?
```
jcoyne: Does this do anything?

mjgiarlo: Yeah, I'm not quite getting what the intent is here.  @cam156

cam156:
@jcoyne @mjgiarlo it seems the representative already gets cleared when destroy is called on a file.  I added a test to be certain this behavior does not change in the future.

jcoyne:
Why did rubocop suddenly start complaining about stuff you didn't add? New release?

mjgiarlo:
That is possible, @jcoyne. We were based on 0.33.0 and 0.34.0 was released on September 5th. We may want to think about pinning the version of rubocop if we find this behavior annoying.

PR #334, opened 2015-09-14
Use a distributed lock to for appending files

jcoyne:
This ensures only one process is updating the linked list at one time.


curation_concerns-models/app/services/curation_concerns/lock_manager.rb
```diff
@@ -0,0 +1,40 @@
+require 'redlock'
+module CurationConcerns
+  class LockManager
+    class UnableToAcquireLockError < StandardError; end
+
+    TIME_TO_HOLD_LOCK = 60_000 # milliseconds
```
terrellt: Can we move these to CC config and document what they do?

PR #335, opened 2015-09-14
Add a way to add more response types.

jcoyne:
For example if you want to add an XML response type, you can now just
override additional_response_formats. This pattern is copied from
Blacklight:
https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/catalog.rb#L143-L168


grosscol:
Looks good to me.  Will be useful if anyone wants a csv, yaml, or other response format.

PR #347, opened 2015-09-29
Create derivatives from the original

jcoyne:
Not from a copy retrieved from the repository
This makes importing faster as less network traffic is required.


jcoyne:
Please don't merge this yet.  Still need to figure out the best way to deal with derivative generation.

jcoyne:
Ready to merge.

PR #350, opened 2015-09-29
Characterization 139

grosscol:
Will squash commits prior to merge.

This PR is pending projecthydra-labs/hydra-works#209


Gemfile
```diff
@@ -7,6 +7,8 @@ gem 'slop', '~> 3.6.0' # This just helps us generate a valid Gemfile.lock when R
 
 gem 'curation_concerns-models', path: './curation_concerns-models'
 
+gem 'hydra-works', git: 'https://github.com/grosscol/hydra-works.git', branch: 'characterization'
```
jcoyne: Can you switch this to something in projecthydra?

curation_concerns-models/app/jobs/characterize_job.rb
```diff
@@ -1,9 +1,12 @@
 class CharacterizeJob < ActiveFedoraIdBasedJob
   queue_as :characterize
 
+  # Calls characterization, saves the generic_file, and queues a job to create derivatives.
+  # Sets original_name property to be same as that of original_file thus over riding characterization.
   def perform(id)
     @id = id
-    CurationConcerns::CharacterizationService.run(generic_file)
```
jcoyne: This change is at cross purposes with #347 What's the best way for us to resolve that?

Gemfile
grosscol: Yes, that's part of the UMICH CONFIG remove this when you squash commit.

curation_concerns-models/app/jobs/characterize_job.rb
grosscol: I'll refactor it.  I'll have to do it anyway once the characterization is done in hydra works.

PR #352, opened 2015-09-30
Move VisibilityCopyJob to curation_concerns-models

jcoyne:
It is referenced by MananagesEmbargoesActor that is also in that gem


jcoyne:
I don't think the Job should be called by the actor, so lets hold off on this for now. See https://github.com/projecthydra-labs/curation_concerns/issues/353

PR #357, opened 2015-10-01
Allow visibility to be set on a collection

jcoyne:


app/forms/curation_concerns/forms/collection_edit_form.rb
```diff
@@ -18,6 +18,15 @@ def select_files
         Hash[all_files]
       end
 
+      class << self
+        # Presntly we can't put visibility in the terms, because the superclass will try to do
```
terrellt: Presently*

jcoyne: Fixed on master. Thanks.

PR #363, opened 2015-10-05
Ask if the Files' visibility should be updated

jcoyne:
Previously Leases were updated automatically. Embargoes were asking if
you wanted to update the files, but only if the embargo had lapsed.
This simplifies the situation by treating Leases and Embargoes
identically and always asking if the visibility should be copied if they
have attached Files.

### DIFF
--- a/app/controllers/concerns/curation_concerns/embargoes_controller_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/embargoes_controller_behavior.rb
@@ -14,10 +14,9 @@ module CurationConcerns
 
     # Removes a single embargo
     def destroy
-      update_files = !curation_concern.under_embargo? # embargo expired
       EmbargoActor.new(curation_concern).destroy
       flash[:notice] = curation_concern.embargo_history.last
-      if update_files
+      if curation_concern.works_generic_work? && curation_concern.generic_files.present?
         redirect_to confirm_curation_concerns_permission_path(curation_concern)
       else
         redirect_to edit_embargo_path(curation_concern)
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/curation_concerns/embargoes_controller_behavior.rb
```diff
@@ -14,10 +14,9 @@ def index
 
     # Removes a single embargo
     def destroy
-      update_files = !curation_concern.under_embargo? # embargo expired
       EmbargoActor.new(curation_concern).destroy
       flash[:notice] = curation_concern.embargo_history.last
-      if update_files
+      if curation_concern.works_generic_work? && curation_concern.generic_files.present?
```
terrellt: This is used a couple times - can we extract this to a method on curation_concern?

jcoyne: In this case `curation_concern` could be a Work or a FileSet (fka GenericFile).  I'm not sure we really want to put it on FileSet, do we?

mjgiarlo: I defer to @terrellt

grosscol: You'd still have to define the method on both of the potential recievers, GenericWork and FileSet.

jcoyne: yea, but I doubt it's used anywhere else for FileSet, so still worth it?

grosscol: I'd rather leave it how it is.  There could be a ticket for DRYing up this at a later date.  It would add a method to FileSet that has no other use case currently.  @terrellt can we punt on this line?

terrellt: :+1: Punt away.

PR #367, opened 2015-10-06
Works and FileSets

jcoyne:


jcoyne:
This is ready for review.

PR #372, opened 2015-10-07
Set minimum version of hydra-head to 9.3.0

jcoyne:
Othewise visibility is always marked as changed on embargos/leases
https://github.com/projecthydra/hydra-head/pull/285

### DIFF
--- a/app/controllers/concerns/curation_concerns/embargoes_controller_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/embargoes_controller_behavior.rb
@@ -9,7 +9,7 @@ module CurationConcerns
     end
 
     def index
-      authorize! :discover, Hydra::AccessControls::Embargo
+      authorize! :index, Hydra::AccessControls::Embargo
     end
 
     # Removes a single embargo
--- a/curation_concerns-models/app/models/concerns/curation_concerns/ability.rb
+++ b/curation_concerns-models/app/models/concerns/curation_concerns/ability.rb
@@ -13,7 +13,12 @@ module CurationConcerns
       # user can version if they can edit
       alias_action :versions, to: :update
 
-      admin_permissions if admin?
+      if admin?
+        admin_permissions
+      else
+        cannot :index, Hydra::AccessControls::Embargo
+        cannot :index, Hydra::AccessControls::Lease
+      end
 
       can :collect, :all
     end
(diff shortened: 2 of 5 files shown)


curation_concerns-models/app/models/concerns/curation_concerns/ability.rb
```diff
@@ -13,7 +13,12 @@ def curation_concerns_permissions
       # user can version if they can edit
       alias_action :versions, to: :update
 
-      admin_permissions if admin?
+      if admin?
+        admin_permissions
+      else
+        cannot :index, Hydra::AccessControls::Embargo
```
mjgiarlo: Why are the explicit `cannot`s needed here?

spec/abilities/embargo_and_lease_ability_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+require 'cancan/matchers'
+
+describe "Ability on embargos and leases" do
+  subject { Ability.new(current_user) }
+
+  let(:current_user) { create(:user) }
+
+  describe "a regular user" do
+    it do
```
mjgiarlo: `it do` looks hella weird.

app/controllers/concerns/curation_concerns/embargoes_controller_behavior.rb
```diff
@@ -9,7 +9,7 @@ module EmbargoesControllerBehavior
     end
 
     def index
-      authorize! :discover, Hydra::AccessControls::Embargo
+      authorize! :index, Hydra::AccessControls::Embargo
```
mjgiarlo: Is the action change just tracking a change in HH?

jcoyne: Yes. hh now grants discover on AF::Base types by default

curation_concerns-models/app/models/concerns/curation_concerns/ability.rb
jcoyne: Because this access exists by default. We have to revoke it

spec/abilities/embargo_and_lease_ability_spec.rb
jcoyne: http://i1.kym-cdn.com/entries/icons/original/000/012/059/they-dont-think-it-be-like-it-is-but-it-do.jpg

mjgiarlo: :laughing:

PR #379, opened 2015-10-08
Implements after_content_create hook

### LINKED ISSUE #378: after_create_content is an invalid configuration option
If you specify a callback for the hook like so:
```ruby
CurationConcerns.configure do |config|
  config.after_create_content = Proc.new { do_something }
end
```
then rails refuses to start because `after_content_create` is not a valid config option. However [these lines](https://github.com/projecthydra-labs/curation_concerns/blob/9d8ebf17b761946ab74576def080504d490d5737/curation_concerns-models/app/jobs/ingest_file_job.rb#L15-L16) suggests that the hook should be there.

afred:
closes #378

### DIFF
--- a/curation_concerns-models/app/jobs/ingest_file_job.rb
+++ b/curation_concerns-models/app/jobs/ingest_file_job.rb
@@ -11,8 +11,6 @@ class IngestFileJob < ActiveJob::Base
     Hydra::Works::UploadFileToFileSet.call(file_set, file, versioning: false)
     file_set.save!
     CurationConcerns::VersioningService.create(file_set.original_file, user_key)
-
-    return unless CurationConcerns.config.respond_to?(:after_create_content)
-    CurationConcerns.config.after_create_content.call(file_set, user_key)
+    CurationConcerns.config.after_create_content.call(file_set, user_key) unless CurationConcerns.config.after_create_content.nil?
   end
 end
--- a/curation_concerns-models/lib/curation_concerns/configuration.rb
+++ b/curation_concerns-models/lib/curation_concerns/configuration.rb
@@ -58,7 +58,8 @@ module CurationConcerns
     attr_accessor :temp_file_base, :enable_local_ingest, :analytic_start_date,
                   :fits_to_desc_mapping, :max_days_between_audits,
                   :resource_types, :resource_types_to_schema,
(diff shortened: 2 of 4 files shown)


jcoyne:
I think the idea here was to subclass `CurationConcerns::Configuration` and implement `after_content_create` as a method rather than set procs.  If you want to change to this approach, we should have the same changes for `after_update_metadata`, `after_destroy`, `after_revert_content`, `after_update_content`, `after_import_local_file_failure`, `after_upload_set_update_success`, `after_upload_set_update_failure`, `after_import_url_failure`, `after_import_url_success`, `after_audit_failure`, etc.

afred:
Subclassing `CurationConcerns::Configuration` does not seem to be the obvious approach. The CurationConcerns installer puts [this file](https://github.com/projecthydra-labs/curation_concerns/blob/5faf2b0d3409ab9843508fa4765433fb31fb9a1d/curation_concerns-models/lib/generators/curation_concerns/models/templates/config/curation_concerns.rb) into your app's initializers. If I'm looking to change the config, then the initializer seems the most logical place to do so.

Also.. if you were to subclass `CurationConcerns::Configruation`, wouldn't you have to override [this method](https://github.com/projecthydra-labs/curation_concerns/blob/ce4e3e5c335c0dc2aff918e1976e58d513f92b58/curation_concerns-models/lib/curation_concerns/configuration.rb#L7-L10), or somehow ensure that `CurationConcerns.config` is getting set before any subsequent calls to `.configure`?

jcoyne:
@afred I took a look at this. It comes from sufia, where Sufia.config is an Engine::Configuration which extends from Railtie::Configuration, which has this method_missing block. https://github.com/rails/rails/blob/9b0b6bf54c83cfdc6329c69571165c34054cea51/railties/lib/rails/railtie/configuration.rb#L89-L95 This allows the config to take arbitrary getters/setters.  So maybe, you'd consider putting that in our config here?  Otherwise, I have no problem with the change you proposed so long as you do it for the other 10 or so hook properties.

afred:
@jcoyne thanks for tracking that down. I will update the PR, but may not get to it until later today or tomorrow.

afred:
tried a new approach in #408. Closing this one

PR #394, opened 2015-10-13
Use an ActiveFedora belongs_to for the representative

jcoyne:


curation_concerns-models/app/models/concerns/curation_concerns/has_representative.rb
```diff
@@ -2,12 +2,8 @@ module CurationConcerns::HasRepresentative
...
   included do
-    property :representative, predicate: RDF::URI.new('http://opaquenamespace.org/ns/hydra/representative'), multiple: false
-  end
-
-  def to_solr(solr_doc = {})
-    super.tap do |doc|
-      doc[Solrizer.solr_name('representative', :stored_searchable)] = representative
-    end
+    belongs_to :representative,
+               predicate: ::RDF::URI('http://opaquenamespace.org/ns/hydra/representative'),
```
terrellt: Is there a better predicate, or do we need to mint this?

jcoyne: This is the same as what we were using before the change.  I don't really care which predicate is used.

grosscol: I couldn't find a more appropriate predicate.  I have an email in to our metadata librarian, but updating the predicate could always be a subsequent issue.

PR #403, opened 2015-10-14
Separate the concept of thumbnail from media

jcoyne:
In certain cases you may want to select one FileSet as the media and a
different FileSet as the thumnail (e.g. audio, video). This provides
that flexibility.


curation_concerns-models/app/models/concerns/curation_concerns/solr_document_behavior.rb
```diff
@@ -41,7 +41,7 @@ def human_readable_type
     end
 
     def representative_id
-      Array(self[Solrizer.solr_name('representative', :symbol)]).first
+      Array(self[Solrizer.solr_name('hasRelatedImage', :symbol)]).first
     end
```
grosscol: Why is this `hasRelated` image and not `hasRelatedMediaFragment` since it is for `representative_id` and not `thumbnail_id`?

jcoyne: Right. Nice catch.

PR #404, opened 2015-10-15
Allow rendered attribute to be a link to facet

jcoyne:
This allows us to have facet links like sufia does
https://github.com/projecthydra/sufia/blob/master/app/views/records/show_fields/_creator.html.erb#L3


jcoyne:
Let's hold off on merging this until #406 is addressed.

njaffer:
@jcoyne, I can work on this if you can provide me some more info. @cam156, I see there is in progress label, which was removed.

PR #407, opened 2015-10-16
Pass object instead of ID to avoid unnecessary lookups

val99erie:


grosscol:
Passing around the `work` instead of the `work_id` makes this code easier to read.

The code looks good.  Appears that the call to `ActiveFedora::Base.find` is moved out of 'attach_file_to_work' and into `process_file`.  Does this save calls to find?

jcoyne:
@grosscol yes, WorkActorBehavior no longer passes an id from an object it already has loaded, only for the Actor to reload it.

PR #408, opened 2015-10-19
Implements a callback api

### LINKED ISSUE #378: after_create_content is an invalid configuration option

afred:
* Allows callbacks for hooks to be defined in configuration block along with
  the rest of config options, for instance those generated by the installer.
* Requires that callback hooks be 'enabled' so that we can all know what hooks
  are available to tie into.

closes #378

### DIFF
--- a/curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb
+++ b/curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb
@@ -73,8 +73,7 @@ module CurationConcerns
       working_file = copy_repository_resource_to_working_directory(file_set)
       make_derivative(file_set.id, working_file)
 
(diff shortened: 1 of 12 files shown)


jcoyne:
Rubocop :facepunch:

jcoyne:
:+1: Looks good to me.

PR #412, opened 2015-10-20
Get the most recent version of the work within the lock

jcoyne:


tpendragon:
Need to rebase to get specs to pass.

spec/actors/curation_concerns/file_set_actor_spec.rb
```diff
@@ -141,4 +141,28 @@
       end
     end
   end
+
+  describe "#attach_file_to_work" do
+    let(:file_set2) { create(:file_set) }
+    let(:file_set3) { create(:file_set) }
+    let!(:work_v1) do
+      work = create(:generic_work)
+      # this is not part of a block on the create, since the work must be saved be fore the representative can be assigned
```
mjgiarlo: s/be fore/before/

spec/actors/curation_concerns/file_set_actor_spec.rb
```diff
@@ -141,4 +141,28 @@
...
+
+    let!(:work_v2) do
+      work = ActiveFedora::Base.find(work_v1.id)
+      work.members << file_set2
+      work.save
+      work
+    end
+
+    it "writes to the most up to date version" do
+      actor.send(:attach_file_to_work, work_v1, file_set3, {})
```
mjgiarlo: Can you add a comment here that `#attach_file_to_work` is private, and that's why you're invoking via `#send`?

spec/actors/curation_concerns/file_set_actor_spec.rb
```diff
@@ -141,4 +141,28 @@
...
+    let(:file_set3) { create(:file_set) }
+    let!(:work_v1) do
+      work = create(:generic_work)
+      # this is not part of a block on the create, since the work must be saved be fore the representative can be assigned
+      work.members << file_set
+      work.save
+      work
+    end
+
+    let!(:work_v2) do
```
mjgiarlo: Can you add a comment here to let folks know that even though `work_v2` is never referenced, this is still a useful (and, in fact, necessary) block of code?

mjgiarlo:
Build failed and was complaining about Redis not having been started. Probably a transient hiccup? Restarted build.

PR #416, opened 2015-10-21
Extract PermissionBadge class

jcoyne:


app/presenters/curation_concerns/permission_badge.rb
```diff
@@ -0,0 +1,61 @@
...
+          'Open Access'
+        elsif registered?
+          I18n.translate('curation_concerns.institution_name')
+        else
+          'Private'
+        end
+      end
+
+      def open_access_with_embargo?
+        if @open_access_with_embargo.nil?
```
mjgiarlo: Can these four lines be replaced with:

```ruby
@open_access_with_embargo ||= open_access? && embargo?
```

app/presenters/curation_concerns/permission_badge.rb
```diff
@@ -0,0 +1,61 @@
...
+          @open_access_with_embargo = open_access? && embargo?
+        end
+        @open_access_with_embargo
+      end
+
+      def open_access?
+        @open_access = @solr_document.visibility == Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC if @open_access.nil?
```
mjgiarlo: Same question as above re: `||=`

app/presenters/curation_concerns/permission_badge.rb
```diff
@@ -0,0 +1,61 @@
...
+        @open_access
+      end
+
+      def registered?
+        @registered = @solr_document.visibility == Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED if @registered.nil?
```
mjgiarlo: Same question as above re: `||=`

mjgiarlo:
Minor questions about how ivars are set in the new class, but meh. Merging.

app/presenters/curation_concerns/permission_badge.rb
jcoyne: No, because `false` doesn't cache in that case.

app/presenters/curation_concerns/permission_badge.rb
jcoyne: nope, `false` would never be cached.

app/presenters/curation_concerns/permission_badge.rb
jcoyne: nope, `false` would never be cached.

PR #418, opened 2015-10-21
update rspec to resolve #373

### LINKED ISSUE #373: Refactor "pending #89" characterization specs
Resolve both `skip 'pending hydra-works#89'` in the file_sets_controller_spec. 

https://github.com/projecthydra-labs/curation_concerns/blob/master/spec/controllers/curation_concerns/file_sets_controller_spec.rb

njaffer:


spec/controllers/curation_concerns/file_sets_controller_spec.rb
```diff
@@ -164,8 +170,6 @@
           expect(CharacterizeJob).to receive(:perform_later).with(file_set.id, kind_of(String))
           post :update, id: file_set, file_set: { files: [file] }
           expect(response).to redirect_to main_app.curation_concerns_file_set_path(file_set)
-          skip 'pending hydra-works#89'
-          expect(file_set.reload.label).to eq 'image.png'
```
mjgiarlo: Can you comment on this deletion? Would this expectation fail?

(To be clear, I'm not even sure why we'd need to check a label here.)

grosscol: I never was able to figure out where label was getting changed.  I think the spirit of that expect was to check that the file_set had actually been updated with the new file.  It seems that this doesn't have to be done by checking `label` but could be done by checking something after reload.

Maybe look at what update should cause to change in the file_set when it updates the files content, and check that it is actually occurring.

njaffer: Yes, the label is nil, so the test fails. @barmintor and I was looking at the implementation and cannot find why the file_set label is being checked here.

jcoyne: I think the label used to get set (before we switched to PCDM).  I don't think it does any longer and I'm not sure that it's important.

PR #425, opened 2015-10-21
Update the solr index of the parent work when thumbnail is created. Story curationexperts/goldenseal#213

val99erie:


jcoyne:
:+1:

PR #434, opened 2015-10-22
Indexing works and file_sets separately instead of lumping them both under file_set_ids_ssim

escowles:


jcoyne:
This is going to cause all the members to load from fedora, a huge penalty. This is why I avoided doing this in the first place.

escowles:
@jcoyne OK, how about just fixing the solr field name to be `member_ids_ssim` instead of `file_set_ids_ssim` then?

tpendragon:
:+1: to member_ids_ssim

PR #441, opened 2015-10-30
Don't prompt to change file visibility when there are no files

jcoyne:


spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -114,9 +114,20 @@
...
+        let(:a_work) { create(:work_with_one_file, user: user) }
+
+        it 'prompts to change the files access' do
+          patch :update, id: a_work
+          expect(response).to redirect_to main_app.confirm_curation_concerns_permission_path(controller.curation_concern)
+        end
+      end
+
+      context 'without children' do
+        it 'prompts to change the files access' do
```
tpendragon: This description isn't right, is it?

PR #445, opened 2015-11-02
Update engine_cart to 0.8.0

jcoyne:


Gemfile
```diff
@@ -17,18 +17,27 @@ group :development, :test do
...
-    gem 'sass-rails', "< 5.0"
-  else
+  if ENV['RAILS_VERSION'].nil? || ENV['RAILS_VERSION'] =~ /^4.2/
+    gem 'responders', "~> 2.0"
     gem 'sass-rails', ">= 5.0"
+  else
+    gem 'sass-rails', "< 5.0"
   end
-
-  extra_file = File.expand_path("../spec/test_app_templates/Gemfile.extra", __FILE__)
```
grosscol: Are we not going to use spec/test_app_templates/Gemfile.extra and instead do the pinning to specific commits in the top level Gemfile?

jcoyne: It's handled here: https://github.com/cbeer/engine_cart/blob/9ab2a6271d6b95c406961be7e03ffb3a764780ef/lib/engine_cart/tasks/engine_cart.rake#L67

PR #450, opened 2015-11-10
Return a single value rather than an array. 

### LINKED ISSUE #449: Mime-type is being returned as an array
The current version of SolrDocumentBehavior returns an array (rather an a single value) for the mime type https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/models/concerns/curation_concerns/solr_document_behavior.rb#L115:

```
    def mime_type
      self[Solrizer.solr_name('mime_type', :stored_sortable)]
    end
```

This causes several methods in HydraWorks to work incorrectly. For example method to detect if a mime type is an office document here https://github.com/projecthydra-labs/hydra-works/blob/master/lib/hydra/works/models/concerns/file_set/mime_types.rb#L22 expects `mime_type` to be a single string like `"text/rtf"` rather than `["text/rtf"]`

I think we should update the code in Curation Concerns to return a single value rather than an array to preserve the semantics that this method used to provide. Otherwise we will need to update everyplace where mime_type is accessed.

hectorcorrea:
Fixes #449

### DIFF
--- a/curation_concerns-models/app/models/concerns/curation_concerns/solr_document_behavior.rb
+++ b/curation_concerns-models/app/models/concerns/curation_concerns/solr_document_behavior.rb
@@ -112,7 +112,7 @@ module CurationConcerns
     end
 
     def mime_type
-      self[Solrizer.solr_name('mime_type', :stored_sortable)]
+      Array(self[Solrizer.solr_name('mime_type', :stored_sortable)]).first
     end
 
     def read_groups


jcoyne:
How is it that a stored sortable (singular) is ever an array?

jcoyne:
Should this have a test?

hectorcorrea:
@jcoyne I can add a test.

As far as the stored_sortable, I left that as it was (which is a change on how Sufia used to store it.) I could change that as well but I wasn't sure about this. Thoughts?

jcoyne:
I would think that solr returns _ssi fields as singular, so that casting to array and taking the first element would be unnecessary.

hectorcorrea:
Nevermind.

PR #460, opened 2015-11-13
Handle ingest of File from IngestLocalFileJob

### LINKED ISSUE #456: IngestLocalFile job attempts to use non-existant method original_filename
```
    NoMethodError:
       undefined method `original_filename' for #<File:spec/mock_upload_directory/world.png>
     # /Users/justin/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/curation_concerns-02e5526e70bf/curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb:114:in `copy_file_to_working_directory'
     # /Users/justin/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/curation_concerns-02e5526e70bf/curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb:59:in `create_content'
     # /Users/justin/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/bundler/gems/curation_concerns-02e5526e70bf/curation_concerns-models/app/jobs/ingest_local_file_job.rb:21:in `perform'
```

grosscol:
Allow use of File in FileSetActor create_content.
Don't depend on original_filename and content_type.
Use AddFileToFileSet to avoid extra calls to save from IngestFileJob.
Update yard docs to indicate File as valid param type.
closes #456

### DIFF
--- a/curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb
+++ b/curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb
@@ -46,14 +46,18 @@ module CurationConcerns
     # Simultaneously moving a preservation copy to the repostiory.
     # TODO: create a job to monitor this directory and prune old files that
     # have made it to the repo
-    # @param [ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user.
+    # @param [File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user.
     def create_content(file)
-      file_set.label ||= file.original_filename
+      # Assign label and title of File Set is necessary.
+      file_set.label ||= file.respond_to?(:original_filename) ? file.original_filename : ::File.basename(file)
       file_set.title = [file_set.label] if file_set.title.blank?
+
+      # Need to save the file_set in order to get an id
       return false unless file_set.save
 
       working_file = copy_file_to_working_directory(file, file_set.id)
-      IngestFileJob.perform_later(file_set.id, working_file, file.content_type, user.user_key)
+      mime_type = file.respond_to?(:content_type) ? file.content_type : nil
+      IngestFileJob.perform_later(file_set.id, working_file, mime_type, user.user_key)
       make_derivative(file_set.id, working_file)
       true
     end
@@ -103,11 +107,13 @@ module CurationConcerns
         CharacterizeJob.perform_later(file_set_id, working_file)
       end
 
-      # @param [ActionDispatch::Http::UploadedFile] file
+      # @param [File, ActionDispatch::Http::UploadedFile] file
       # @param [String] id the identifer
       # @return [String] path of the working file
       def copy_file_to_working_directory(file, id)
-        copy_stream_to_working_directory(id, file.original_filename, file)
+        # file_set.label not gaurunteed to be set at this point (e.g. if called from update_content)
+        file_set.label ||= file.respond_to?(:original_filename) ? file.original_filename : ::File.basename(file)
+        copy_stream_to_working_directory(id, file_set.label, file)
       end
 
       # @param [FileSet] file_set the resource
--- a/curation_concerns-models/app/jobs/ingest_file_job.rb
+++ b/curation_concerns-models/app/jobs/ingest_file_job.rb
@@ -3,13 +3,23 @@ class IngestFileJob < ActiveJob::Base
 
   def perform(file_set_id, filename, mime_type, user_key)
     file_set = FileSet.find(file_set_id)
-    file = Hydra::Derivatives::IoDecorator.new(File.open(filename, "rb"))
-    file.mime_type = mime_type
-    file.original_name = File.basename(filename)
 
-    # Tell UploadFileToGenericFile service to skip versioning because versions will be minted by VersionCommitter (called by save_characterize_and_record_committer) when necessary
-    Hydra::Works::UploadFileToFileSet.call(file_set, file, versioning: false)
+    file = File.open(filename, "rb")
+    # If mime-type is known, wrap in an IO decorator
+    # Otherwise allow Hydra::Works service to determine mime_type
+    if mime_type
+      file = Hydra::Derivatives::IoDecorator.new(file)
+      file.mime_type = mime_type
+      file.original_name = File.basename(filename)
+    end
+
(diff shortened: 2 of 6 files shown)


curation_concerns-models/app/jobs/ingest_file_job.rb
```diff
@@ -3,13 +3,27 @@ class IngestFileJob < ActiveJob::Base
...
+      file.mime_type = mime_type
+      file.original_name = File.basename(filename)
+    end
+
+    # Tell AddFileToFileSet service to skip versioning because versions will be minted by VersionCommitter (called by save_characterize_and_record_committer) when necessary
+    Hydra::Works::AddFileToFileSet.call(file_set, file, :original_file, versioning: false)
+
+    # Assign label and title of File Set after original_name has been determined.
+    file_set.label ||= file_set.original_file.original_name
+    file_set.title = [file_set.label] if file_set.title.blank?
```
grosscol: Setting file_set.title here does not get persisted according to the specs, and I'm baffled as to why that is the case.

curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb
```diff
@@ -50,14 +50,14 @@ def create_metadata(upload_set_id, work, file_set_params = {})
     # Simultaneously moving a preservation copy to the repostiory.
     # TODO: create a job to monitor this directory and prune old files that
     # have made it to the repo
-    # @param [ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user.
+    # @param [File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user.
     def create_content(file)
-      file_set.label ||= file.original_filename
-      file_set.title = [file_set.label] if file_set.title.blank?
```
grosscol: Doing `file_set.title = [file_set.label] if file_set.title.blank?` here satisfies the tests.  Putting the same line in the IngestFileJob does not persist the title.

curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb
```diff
@@ -107,11 +111,12 @@ def make_derivative(file_set_id, working_file)
         CharacterizeJob.perform_later(file_set_id, working_file)
       end
 
-      # @param [ActionDispatch::Http::UploadedFile] file
+      # @param [File, ActionDispatch::Http::UploadedFile] file
       # @param [String] id the identifer
       # @return [String] path of the working file
       def copy_file_to_working_directory(file, id)
-        copy_stream_to_working_directory(id, file.original_filename, file)
+        name = file.respond_to?(:original_filename) ? file.original_filename : ::File.basename(file)
```
jcoyne: why don't we use `file_set.label` here (or pass it as a parameter)?  Then we avoid repeating this test.

jcoyne:
Please add a couple of tests to this PR.

curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb
grosscol: Yes.  I was reluctant to change any function signatures, but I did not notice that this one is marked private.  I'll update this to do the respond_to check once and then pass along the result where needed.

grosscol:
A few tests added.

spec/jobs/ingest_file_job_spec.rb
```diff
@@ -5,11 +5,17 @@
...
   let(:user)     { create(:user) }
 
-  it 'uses the provided mime_type' do
+  it 'uses a provided mime_type' do
     described_class.perform_now(file_set.id, filename, 'image/png', 'bob')
     expect(file_set.reload.original_file.mime_type).to eq 'image/png'
   end
 
+  it 'does not decorate File when not given mime_type' do
+    allow(CurationConcerns::VersioningService).to receive(:create)
```
jcoyne: Why should VersioningService ever be called? wasn't `versioning: false` going to prevent it?

spec/jobs/ingest_file_job_spec.rb
```diff
@@ -5,11 +5,17 @@
   let(:filename) { fixture_file_path('/world.png') }
   let(:user)     { create(:user) }
 
-  it 'uses the provided mime_type' do
+  it 'uses a provided mime_type' do
     described_class.perform_now(file_set.id, filename, 'image/png', 'bob')
     expect(file_set.reload.original_file.mime_type).to eq 'image/png'
   end
 
+  it 'does not decorate File when not given mime_type' do
```
jcoyne: can you make this a context block "when mime_type is not given"

spec/jobs/ingest_file_job_spec.rb
grosscol: The call to `versioning: false` is an instruction to the Hydra::Works machinery so that the CurationConcerns machinery can do its own versioning.

jcoyne: :ok_hand: Thanks for the explanation.

curation_concerns-models/app/actors/curation_concerns/file_set_actor.rb
```diff
@@ -103,11 +107,13 @@ def make_derivative(file_set_id, working_file)
...
       end
 
-      # @param [ActionDispatch::Http::UploadedFile] file
+      # @param [File, ActionDispatch::Http::UploadedFile] file
       # @param [String] id the identifer
       # @return [String] path of the working file
       def copy_file_to_working_directory(file, id)
-        copy_stream_to_working_directory(id, file.original_filename, file)
+        # file_set.label not gaurunteed to be set at this point (e.g. if called from update_content)
+        file_set.label ||= file.respond_to?(:original_filename) ? file.original_filename : ::File.basename(file)
```
jcoyne: label has already been set on line 52. Is there ever a case it gets here without a label set?

jcoyne: I see that it could get called from update_content, but in that case file_set never gets saved, so it's not required, right?

grosscol: The spec seemed to imply that a potential use case was updating a file_set which had no content to begin with so the label would not have already been set.  I assumed the file_set to be saved after the update.

PR #461, opened 2015-11-16
When an unauthorized user does create, show the unauthorized page

jcoyne:
rather than redirecting to the root path


tpendragon:
Spec?

jcoyne:
@tpendragon added

PR #464, opened 2015-11-16
Generate models into CurationConcerns namespace

### LINKED ISSUE #463: Generated models are not in the curation_concerns namespace.
When we generate a model it is in the root namespace (i.e. `GenericWork`).  However the controller we generate is in the `CurationConcerns` namespace (i.e. `CurationConcerns::GenericWorksController`).

We should do something to match these into the same namespace.  I propose we have the models generate into the CurationConcerns ns.

jcoyne:
Backwards incompatible. Need to migrate hasModel assertions.
Fixes #463


jcoyne:
@tpendragon @jpstroop @escowles  How awful will it be for Princeton?

tpendragon:
@jcoyne We don't have a production system. `DELETE http://fedora/rest/` is cool with me.

jpstroop:
Not worried about change overhead for Princeton, but if I'm reading the changes correctly (e.g. [`curation_concerns/spec/features/work_generator_spec.rb`](https://github.com/projecthydra-labs/curation_concerns/blob/namespace_models/spec/features/work_generator_spec.rb)), now if I run `rails generate curation_concerns:catapult` we're going to generate `CurationConcerns::Catapult`? That also seems confusing. The name (and namespace) of my model should be up to me. Could the code generated by the installer be in the CC namespace, but anything generated by `rails generate curation_concerns:x` be in the default namespace?

I agree that the inconsistency is a problem; is there something keeping this from going the other way, i.e. take the controllers, actors, forms out of the cc namespace (sorry, it's been a while).

jcoyne:
@jpstroop The problem is that the views we're rendering are all in `app/views/curation_concerns/base` Where would those go?

jpstroop:
Bugger...I guess no solution is going to be perfect. @tpendragon, @escowles, any thoughts?

jcoyne:
The immediate problem I'm trying to solve is this:
```ruby
model = GenericWork.new
model.model_name.singular_route_key
"generic_work"
```
That ought to be `curation_concerns_generic_work`. I can fix that without changing any namespaces, but messing with the namespaces would fix the underlying problem.

jpstroop:
OK, I'm :+1:

jcoyne:
Okay, please hold off on merging this. I have another commit to add.

jcoyne:
I don't like that this has changed the params key (mostly of the json api), so I'm going to try one other thing.

Issue #468, opened 2015-11-18
UploadSet should have a depositor property


jcoyne:
https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/models/concerns/curation_concerns/upload_set_behavior.rb#L10

Use `RDF::Vocab::MARCRelators.dpt` as the predicate

hectorcorrea:
Per @jcoyne: we should also investigate if `creator` is still needed. Not sure they (`creator` vs `depositor`) will ever be different.

hectorcorrea:
@mjgiarlo @jcoyne I get the impression that the `UploadSet.creator` is user defined (see https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/upload_sets/_metadata.html.erb#L35) 

Not sure if this helps to answer the question on whether they will ever be different. Thoughts?

mjgiarlo:
@hectorcorrea The UploadSet **form**'s creator value is user defined, and is applied to all things in the UploadSet, not the UploadSet itself. UploadSets should not have descriptive metadata. I just spun up Sufia (6.x) master and deposited a few files together, and found that the one Batch object in Fedora had no titles and no creators assigned. Can someone else verify this? Now I'm wondering how we ever used these fields, and why they are no longer used. (Did my batch update job misfire?) I think we should first get to the bottom of how we're currently using these fields in Sufia 6.x, and that will help determine what we do, per @hectorcorrea's suggestion.

jcoyne:
@mjgiarlo I'm not sure the values on the batch form are ever persisted on the UploadSet object, I think they're just pushed on the ActiveJob queue.

mjgiarlo:
@jcoyne It's kinda weird, though has never been problematic, that Batches don't have depositor metadata applied -- I'm guessing this changed in the move to Fedora 4. Anyone can theoretically update anyone else's batches if they know the ID. Maybe?

jcoyne:
Yes, very odd. Why does it need a depositor?  BatchUpdateJob in sufia didn't have one.
It didn't have one here on July 15: https://github.com/projecthydra-labs/sufia-core/blob/e478eb243b06cae8f9745352af927263f4bdefdd/sufia-models/app/jobs/batch_update_job.rb

But it gets added here Sept 1:
[...]

PR #471, opened 2015-11-18
UploadSet now has a depositor property.

hectorcorrea:
Addresses #468

### DIFF
--- a/curation_concerns-models/app/jobs/upload_set_update_job.rb
+++ b/curation_concerns-models/app/jobs/upload_set_update_job.rb
@@ -28,16 +28,14 @@ class UploadSetUpdateJob < ActiveJob::Base
     if denied.empty?
       unless saved.empty?
         if CurationConcerns.config.callback.set?(:after_upload_set_update_success)
-          login = upload_set.depositor
-          user = User.find_by_user_key(login)
+          user = User.find_by_user_key(@login)
           CurationConcerns.config.callback.run(:after_upload_set_update_success, user, upload_set, log.created_at)
         end
         return true
       end
     else
       if CurationConcerns.config.callback.set?(:after_upload_set_update_failure)
-        login = upload_set.depositor
-        user = User.find_by_user_key(login)
+        user = User.find_by_user_key(@login)
         CurationConcerns.config.callback.run(:after_upload_set_update_failure. user, upload_set, log.created_at)
       end
       return false
--- a/curation_concerns-models/app/models/concerns/curation_concerns/upload_set_behavior.rb
+++ b/curation_concerns-models/app/models/concerns/curation_concerns/upload_set_behavior.rb
@@ -7,7 +7,6 @@ module CurationConcerns
     included do
       has_many :file_sets, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf
 
-      property :creator, predicate: ::RDF::DC.creator
       property :title, predicate: ::RDF::DC.title
       property :status, predicate: ::RDF::DC.type
     end
--- a/spec/models/upload_set_spec.rb
+++ b/spec/models/upload_set_spec.rb
@@ -2,11 +2,10 @@ require 'spec_helper'
 
 describe UploadSet do
   let(:user) { FactoryGirl.find_or_create(:jill) }
-  let(:upload_set) { described_class.create(title: ["test collection"], creator: [user.user_key]) }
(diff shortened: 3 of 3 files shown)


hectorcorrea:
Should I add the new `depositor` field to the UploadSet forms so that the user can set/change its value or is the expectation that the applications will initialize this value with the current user?

jcoyne:
I'm not sure adding depositor is required, see updates on #468

hectorcorrea:
Updated with changes suggested on #468

jcoyne:
:+1:

curation_concerns-models/app/models/concerns/curation_concerns/upload_set_behavior.rb
```diff
@@ -7,7 +7,6 @@ module UploadSetBehavior
     included do
       has_many :file_sets, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf
 
-      property :creator, predicate: ::RDF::DC.creator
```
mjgiarlo: :+1: and I think we should replace with a marcrel-based depositor (and use the apply_depositor_metadata method on creation of UploadSets)

mjgiarlo:
OK, left comments on #468.

curation_concerns-models/app/models/concerns/curation_concerns/upload_set_behavior.rb
jcoyne: apply_depositor_metadata persists associated web acls into the repo. This will slow things down. Are you sure we want that?

mjgiarlo: What will slow things down? **Writing** those ACLs or their presence?

It seems broken to me that we have all these UploadSet objects just chilling out in Fedora without any association between them and their owners. Maybe this is an orthogonal concern that should be ticketed and discussed separately with more people?

jcoyne: There are associations between them and the files that are uploaded.  The files that are uploaded have the owners.  Can we infer their ownership through that?  UploadSets have no metadata, and are not editable, or deletable, they're basically just empty resources, used for a URL, right?  Why do we need to make them owned by anyone?

mjgiarlo: We can *kinda* infer ownership through the `belongs_to` relationship declared on FileSets, though proxy depositor/transfer functionality could introduce inconsistencies. If I deposit 10 files and then transfer n of them to someone else at some point, who gets to perform operations on the UploadSet? Me or someone else? Part of this is that I think having the UploadSets persist in the repository is a potential place to hang other functionality in the future. Perhaps that is YAGNI territory, and we could put a flag in this for figuring out at that point.

jcoyne: Sufia 6 is getting by just fine without, so I'd suggest not letting this PR get blocked.

mjgiarlo: I'm cool with that. I still have a question about wiping out the callback hooks above, which Sufia already uses to send notifications.

curation_concerns-models/app/jobs/upload_set_update_job.rb
```diff
@@ -26,20 +26,8 @@ def perform(login, upload_set_id, title, file_attributes, visibility)
     upload_set.update(status: ["Complete"])
 
     if denied.empty?
-      unless saved.empty?
```
mjgiarlo: I don't think we should wipe out the blocks of code in this method. Or, at very least, if we do, we should ticket adding them back. We use these callbacks in Sufia to send notifications.

jcoyne: Ah, good catch, I didn't see that these are the descendants of `send_user_success_message` and `send_user_failure_message`

jcoyne: I think we can restore the blocks and just use the `@login` rather than the depositor.

mjgiarlo: :ok: #worksforme

hectorcorrea:
I've restored the after hook events and updated them to use `@login` rather than the (now gone) `depositor`

PR #472, opened 2015-11-19
Allow sort_and_per_page to be used for edit

jcoyne:
CurationConcerns currently only uses sort_and_per_page on the
Collection's show view. Sufia also uses it on the edit view. On the edit
view we use `@form` rather than `@presenter`, so replace the ivars with
local variables.


tpendragon:
Rubocop.

jcoyne:
@tpendragon all better

PR #477, opened 2015-11-20
Call callbacks with two arguments.

jcoyne:


mjgiarlo:
:+1: for me. I'd like one of @tpendragon @escowles @tpendragon @tampakis to give the :ok: before shipping.

jpstroop:
:+1: Fine here.

mjgiarlo:
Thank you, Stroop!

PR #484, opened 2015-11-25
Add a read-only audit service method

jcoyne:
The existing method kicks off audits if they are needed. This adds a way
to just get the status without triggering new audits.  Ported from:
https://github.com/projecthydra/sufia/pull/1178

### DIFF
--- a/curation_concerns-models/app/services/curation_concerns/file_set_audit_service.rb
+++ b/curation_concerns-models/app/services/curation_concerns/file_set_audit_service.rb
@@ -1,16 +1,31 @@
 module CurationConcerns
   class FileSetAuditService
     attr_reader :file_set
-    def initialize(file)
-      @file_set = file
+    def initialize(file_set)
+      @file_set = file_set
     end
 
     NO_RUNS = 999
 
     # provides a human readable version of the audit status
+    # This may trigger audits to be run if required
     # @param [Hydra::PCDM::File] file the file to get the audit status for, defaults to the original_file.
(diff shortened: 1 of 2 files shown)


curation_concerns-models/app/services/curation_concerns/file_set_audit_service.rb
```diff
@@ -1,18 +1,33 @@
...
     # @param [Hydra::PCDM::File] file the file to get the audit status for, defaults to the original_file.
     def human_readable_audit_status(file = file_set.original_file)
       audit_stat(file)
     end
 
+    # Check the file by only what is in the audit log.
+    # Do not try to access the versions if we do not have access to them.
+    # Use this when a file_set is loaded from solr instead of fedora
+    def logged_audit_status
+      audit_results = ChecksumAuditLog.logs_for(file_set.id, "original_file")
```
tpendragon: Should this accept an argument? Hard-coded original_file isn't a problem? (Not totally sure what the audit service stuff does, so it may not be.)

jcoyne: YAGNI?  This is the way it is in Sufia. Currently I don't have a case where I'd need something else, so why change it until we find that case?

grosscol: I don't have a problem hard coding 'original_file' for the time being.  It is the default value in the hydra works code, and this far up in the stack I wouldn't expect anyone to want to be using any other value at this time.

PR #489, opened 2015-11-30
Use NullLogger in case ActiveFedora::Base.logger is nil

awead:
`ActiveFedora::Base.logger` can be nil, so provide a null object to stand in its place that accepts all the same methods as a logger, but just returns nil. This has additional uses in other dependent gems that might send logging messages to nil objects such as `Rails.logger`.

### DIFF
--- a/spec/lib/curation_concerns/null_logger_spec.rb
+++ b/spec/lib/curation_concerns/null_logger_spec.rb
@@ -1,0 +1,10 @@
+require 'spec_helper'
+
+describe CurationConcerns::NullLogger do
+  subject { described_class.new }
+  its(:debug) { is_expected.to be_nil }
+  its(:info) { is_expected.to be_nil }
+  its(:warn) { is_expected.to be_nil }
+  its(:error) { is_expected.to be_nil }
+  its(:fatal) { is_expected.to be_nil }
+end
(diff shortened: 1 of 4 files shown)


spec/lib/curation_concerns/null_logger_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'spec_helper'
+
+describe CurationConcerns::NullLogger do
+  subject { described_class.new }
+  context "with an info message" do
```
mjgiarlo: I'm not sure what this `context` block adds and I find it a bit confusing -- I wouldn't call `logger.error` to be an "info" message. But I dig the PR and I wouldn't hold it up over this.

mjgiarlo:
:+1: if Travis is :green_heart: 

Thoughts, @escowles @tpendragon @tampakis @jpstroop @jcoyne ?

awead:
@mjgiarlo ah, yeah, that context block didn't make sense anymore. I took it out.

awead:
I made `logger` private in the audit job because coveralls was complaining.

awead:
My guess about Coveralls was incorrect. Coveralls is complaining about lines that aren't affected by this PR: https://coveralls.io/builds/4302121/source?filename=curation_concerns-models%2Fapp%2Fjobs%2Faudit_job.rb#L22

PR #503, opened 2015-12-02
Delegate title_or_label from the FileSetPresenter to the SolrDocument

hectorcorrea:
@awead As you suggested, here is the PR with some of the missing properties in the presenter.

Adds `title_or_label` to the presenter.

Document in the test other properties that should also be delegated. See TODO in the test.

Waiting on feedback from the team at the daily.

### DIFF
--- a/spec/presenters/curation_concerns/file_set_presenter_spec.rb
+++ b/spec/presenters/curation_concerns/file_set_presenter_spec.rb
@@ -43,24 +43,17 @@ describe CurationConcerns::FileSetPresenter do
     it { is_expected.to eq 'File Set' }
   end
 
-  describe "date_uploaded" do
-    it "delegates to the solr_document" do
-      expect(solr_document).to receive(:date_uploaded)
-      presenter.date_uploaded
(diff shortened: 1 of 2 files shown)


spec/presenters/curation_concerns/file_set_presenter_spec.rb
```diff
@@ -43,24 +43,22 @@
     it { is_expected.to eq 'File Set' }
   end
 
-  describe "date_uploaded" do
-    it "delegates to the solr_document" do
-      expect(solr_document).to receive(:date_uploaded)
-      presenter.date_uploaded
+  describe "properties delegated to solr_document" do
+    let(:solr_properties) do
+      # TODO: These properties are still missing
```
jcoyne: I don't think this is helpful here. Can we just say "No" for now?  We've not demonstrated a need at this time.

jcoyne:
Squash commits please.

hectorcorrea:
Squashed commits and removed TODO from comments.

mjgiarlo:
:+1:

PR #507, opened 2015-12-02
DON'T MERGE: Fixes the link being wider than the image

### LINKED ISSUE #496: Port image display change
https://github.com/projecthydra/sufia/commit/b4a738a608f76d7392b2b7505b8d7b1278894e1f

awead:
Fixes the link being wider than the image if the browser window is expanded. In order to just have the link wrap the intrinsic value of the image, switch img-responsive to inline block. Fixes #496


awead:
I'm not sure we want this port. With the unmodified current app, it looks like:
current
With this commit, it looks like:
override

hectorcorrea:
Accidentally merged this PR even though it says DON'T MERGE on the subject :( . Reverted the merge via PR #511

PR #538, opened 2015-12-07
Add file_format() to FileSetIndexer. Fixes #263

### LINKED ISSUE #263: Move file_format() into a CurationConcerns::FileSetIndexer
https://github.com/projecthydra-labs/hydra-works/blob/master/lib/hydra/works/models/concerns/file_set/mime_types.rb#L31-L39

It's not used in hydra-works at all. It's only ever called by the indexer.

jcoyne:


mjgiarlo:
:+1:

PR #544, opened 2015-12-08
UploadSets collect works rather than FileSets

### LINKED ISSUE #541: Move CurationConcerns::FileSet::BelongsToUploadSets to GenericWork
This is so you can set an upload_set_id on a work for a file that you've uploaded.

jcoyne:
Fixes #541


curation_concerns-models/app/models/concerns/curation_concerns/upload_set_behavior.rb
```diff
@@ -5,7 +5,7 @@ module UploadSetBehavior
     include CurationConcerns::Noid
 
     included do
-      has_many :file_sets, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf
+      has_many :works, predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf, class_name: 'ActiveFedora::Base'
```
awead: do we need to specify a class_name here?

jcoyne: Yes. If you don't it attempts to load classes named `Work`, which doesn't exist.

Issue #555, opened 2015-12-14
Error on advanced search page


awead:
This is also happening in projecthydra/sufia#1450

```
RuntimeError - current_ability has not been set on #<CurationConcerns::SearchBuilder:0x007fe9ddd420a0>:
  hydra-access-controls (9.5.0) lib/hydra/access_controls_enforcement.rb:19:in `current_ability'
  /Users/awead/Projects/Github/projecthydra-labs/curation_concerns/app/search_builders/curation_concerns/search_builder.rb:15:in `gated_discovery_filters'
  hydra-access-controls (9.5.0) lib/hydra/access_controls_enforcement.rb:91:in `apply_gated_discovery'
  hydra-access-controls (9.5.0) lib/hydra/access_controls_enforcement.rb:73:in `add_access_controls_to_solr_params'
  blacklight (5.17.0) lib/blacklight/search_builder.rb:150:in `block (2 levels) in processed_parameters'
  blacklight (5.17.0) lib/blacklight/search_builder.rb:145:in `block in processed_parameters'
  blacklight (5.17.0) lib/blacklight/search_builder.rb:144:in `processed_parameters'
  blacklight (5.17.0) lib/blacklight/search_builder.rb:116:in `to_hash'
  blacklight (5.17.0) lib/blacklight/solr/repository.rb:44:in `block in send_and_receive'
  activesupport (4.2.5) lib/active_support/benchmarkable.rb:41:in `block in benchmark'
  activesupport (4.2.5) lib/active_support/core_ext/benchmark.rb:12:in `block in ms'
  /Users/awead/.rubies/ruby-2.2.3/lib/ruby/2.2.0/benchmark.rb:303:in `realtime'
  activesupport (4.2.5) lib/active_support/core_ext/benchmark.rb:12:in `ms'
  activesupport (4.2.5) lib/active_support/benchmarkable.rb:41:in `benchmark'
  blacklight (5.17.0) lib/blacklight/solr/repository.rb:42:in `send_and_receive'
  blacklight (5.17.0) lib/blacklight/solr/repository.rb:28:in `search'
  blacklight (5.17.0) app/controllers/concerns/blacklight/search_helper.rb:109:in `search_results'
  blacklight_advanced_search (5.2.1) app/controllers/blacklight_advanced_search/advanced_controller.rb:16:in `get_advanced_search_facets'
[...]

PR #566, opened 2015-12-19
fix advanced search ability issue

njaffer:
related to #555 . The current_user and ability is nil.


jcoyne:
I don't think this is the right fix. Rather than handling a nil current_ability, current_abolity should never be allowed to be nil.

jcoyne:
In other words, the fix we're looking for is to alter the controllers so that they pass the current_ability into the search builder.

njaffer:
@jcoyne , I am looking at the code and when the HydraAccess is called from CatalogController#index, the @current_ability is not nil, but when it is called from AdvancedController#index, the @current_ability is nil

jcoyne:
@njaffer that's right.  When you are done with this ticket, AdvancedController#index should pass the current_ability to the SearchBuilder and it will not be nil.

PR #577, opened 2016-01-07
Small refactor to allow easily overriding the behavior after a FileSet is created

escowles:
See https://github.com/pulibrary/plum/pull/318


jcoyne:
I like this, but why not add `after_update_destination` too?

PR #578, opened 2016-01-09
Delete derivatives when FileSet is destroyed.

tpendragon:


curation_concerns-models/app/services/curation_concerns/derivative_path.rb
```diff
@@ -7,8 +7,19 @@ def derivative_path_for_reference(object, destination_name)
...
+      # Returns all derivatives on the filesystem for this object.
+      def derivatives_for_reference(object)
+        Dir.glob(root_path(object).join("*")).select do |path|
+          path.start_with?(derivative_path(object, "", ""))
+        end
+      end
+
       private
 
+        def root_path(object)
```
jcoyne: Can you add a comment about what this method does?

curation_concerns-models/app/services/curation_concerns/derivative_path.rb
```diff
@@ -7,8 +7,19 @@ def derivative_path_for_reference(object, destination_name)
         derivative_path(object, extension_for(destination_name), destination_name)
       end
 
+      # Returns all derivatives on the filesystem for this object.
```
jcoyne: The derivatives themselves, or a list of paths?

curation_concerns-models/app/services/curation_concerns/derivative_path.rb
```diff
@@ -7,8 +7,19 @@ def derivative_path_for_reference(object, destination_name)
...
+      def derivatives_for_reference(object)
+        Dir.glob(root_path(object).join("*")).select do |path|
+          path.start_with?(derivative_path(object, "", ""))
```
jcoyne: can `derivative_path(object, "", "")` be factored out.

curation_concerns-models/app/services/curation_concerns/derivative_path.rb
```diff
@@ -7,8 +7,19 @@ def derivative_path_for_reference(object, destination_name)
...
+        end
+      end
+
       private
 
+        def root_path(object)
+          Pathname.new(derivative_path(object, "", "")).dirname
```
jcoyne: here's the other place we're using `derivative_path(object, "", "")`

curation_concerns-models/app/services/curation_concerns/derivative_path.rb
grosscol: What do you think about having `""` be the default value for those two args in `def derivative_path` ?

tpendragon:
@jcoyne Take another look?

PR #581, opened 2016-01-11
Make attribute_renderer#label public

tpendragon:
So that other render methods can generate a field label, even if it's
not HTML.


app/renderers/curation_concerns/attribute_renderer.rb
```diff
@@ -32,13 +32,13 @@ def render
       markup.html_safe
     end
 
-    private
+    def label
```
jcoyne: Add a comment on what this method is for. Note that it could be overridden (so we don't make it private again)

PR #590, opened 2016-01-25
Specify full path to constant.

jcoyne:
If an application has FileSetIndexer defined, a `NameError` could be raised:

```
     NameError:
       uninitialized constant FileSetIndexer::STORED_INTEGER
```


mjgiarlo:
Travis = :red_circle:

jcoyne:
Travis is timing out on the jetty downla

mjgiarlo:
maybe we should use git submodules

hahahaha

PR #598, opened 2016-01-29
Fixing bug in looking up parent_id

escowles:
* using existing `lookup_parent_from_child` method to lookup parent instead of duplicating


jcoyne:
:+1:

PR #603, opened 2016-02-04
Use absolute path to CatalogController.

tpendragon:


jcoyne:
:+1:

jcoyne:
This looks like a duplicate of #601

PR #607, opened 2016-02-06
Update HydraWorks dependency

grosscol:
Update hydra works dependency

PR #608, opened 2016-02-09
Remove GenericWork from Solr index

grosscol:
Removing work from solr index before FileSets are destroyed so in_objects lookup  does not explode.  This fixed one of two errors that arise from including hydra-works 0.7.0


jcoyne:
:+1: @tpendragon do you have an opinion?

grosscol:
Rebased, but fix for error depends on #608.  I would have done both in the same PR, but the removing a work from the solr index before deleting it's files seemed sufficiently different to warrant it's own examination.

jcoyne:
@grosscol was that last comment supposed to go on a different ticket, or were you intending to reference a different ticket?

grosscol:
@jcoyne yeah.  Switched tabs and didn't notice.  That's the comment that should be on #607.

tpendragon:
Seems fine to me

PR #615, opened 2016-02-12
change @form to curation_concerns

### LINKED ISSUE #614: render after validation failure leads to error
`app/controllers/concerns/curation_concerns/curation_concern_controller.rb:39`
which is the call to `render 'new'`
```
      setup_form
      respond_to do |wants|
        wants.html { render 'new', status: :unprocessable_entity }
```
The reported error is: `undefined method `model_name' for nil:NilClass`

njaffer:
closes #614

### DIFF
--- a/app/views/curation_concerns/base/_form.html.erb
+++ b/app/views/curation_concerns/base/_form.html.erb
@@ -1,4 +1,4 @@
-<%= simple_form_for [main_app, @form],
+<%= simple_form_for [main_app, curation_concern],
                     wrapper_mappings: { multifile: :horizontal_file_input } do |f| %>
   <% if f.error_notification -%>
     <div class="alert alert-danger fade in">


njaffer:
The form validation was not working after this commit: https://github.com/projecthydra-labs/curation_concerns/commit/4664a64858704a92736c7ddf4d62f1a6faf4e04e#diff-687690e9fb9bbba7f86082293c1544afL1

jcoyne:
I think we should figure out why the @form object isn't working and fix that rather than revert the commit.

jcoyne:
@njaffer It might be helpful if you could first write a test that elucidates the failure before fixing it. That would help me see what's going wrong.

njaffer:
@jcoyne, can you please point me to the test that you had for the change to @form from curation_concern? It would be helpful to see why the test did not fail when the validation of required fields are not working with @form

jcoyne:
The only test that was added was for the `new` method.  The create with failure was not tested:  https://github.com/projecthydra-labs/curation_concerns/commit/4664a64858704a92736c7ddf4d62f1a6faf4e04e#diff-6ecde689350b1846fb9ba6acf8b630afR49

I'm working on adding a test for that now.

jcoyne:
Here's the controller test that shows where the form isn't created:
```ruby
context 'when create fails' do
   it 'draws the form again' do
    expect_any_instance_of(CurationConcerns::GenericWorkActor).to receive(:create).and_return(false)
    post :create, generic_work: { title: ['a title'] }
    expect(response.status).to eq 422
    expect(assigns[:form]).to be_kind_of CurationConcerns::GenericWorkForm
  end
end
```

jcoyne:
Here's my take on how we fix this issue: https://github.com/projecthydra-labs/curation_concerns/pull/623

njaffer:
#623 over rides this PR

PR #620, opened 2016-02-16
Don't consider the :q parameter on the collection show page. Fixes #619

### LINKED ISSUE #619: If the `:q` parameter is set on the collection show page, you get "not authorized"

jcoyne:


jcoyne:
Blocked by #618 (which fixes the build)

PR #612, opened 2016-02-11
Implement File Manager UI

tpendragon:
Allows for reordering Works and assigning labels to them in one
interface.

PR #625, opened 2016-02-19
remove ordered_member_ids from work form

njaffer:
fixes the error: "GenericWork does not have an attribute `ordered_member_ids’"

This might not be the right solution. The error is generated in Sufia, when a new work is created. Here are the steps to generate the error:
1. Create a new work.
2. Hit Save
The error says: ActiveFedora::UnknownAttributeError Exception: GenericWork does not have an attribute `ordered_member_ids’
In the process of debugging, I noticed that when I create a form using CC, the attributes in base actor are:
{"title"=>[], "creator"=>[], "contributor"=>[], "description"=>[], "rights"=>[], "publisher"=>[], "subject"=>[], "language"=>[], "visibility"=>"restricted”}
But when base actor is called from Sufia, the attributes are:
(byebug) attributes
{"title"=>[], "creator"=>[], "contributor"=>[], "description"=>[], "tag"=>[], "publisher"=>[], "date_created"=>[], "subject"=>[], "language"=>[], "identifier"=>[], "based_near"=>[], "related_url"=>[], "visibility"=>"restricted", "ordered_member_ids"=>[], "resource_type"=>[]}

And that is where the error is generated..
If I remove the :ordered_member_ids from work_form.rb, I can save the new work form  in CC and Sufia

Any thoughts? @mjgiarlo , @grosscol,  @tpendragon , @awead , @jcoyne

### DIFF
--- a/app/forms/curation_concerns/forms/work_form.rb
+++ b/app/forms/curation_concerns/forms/work_form.rb
@@ -14,7 +14,7 @@ module CurationConcerns
                     :representative_id, :thumbnail_id, :files,
                     :visibility_during_embargo, :embargo_release_date, :visibility_after_embargo,
                     :visibility_during_lease, :lease_expiration_date, :visibility_after_lease,
-                    :visibility, :ordered_member_ids]
+                    :visibility]
 
       # @param [ActiveFedora::Base,#member_ids] model
       # @param [Ability] current_ability


mjgiarlo:
It looks like @tpendragon added this recently, so I don't know if the right solution is to remove it:

https://github.com/projecthydra-labs/curation_concerns/commit/8f8441c1cb0cd16d7c4114ed2c4e892fd5c636f6#diff-f51c999ab73199d1d7c7d66f9ad87fbe

mjgiarlo:
What version of CC are you pulling in, @njaffer? 0.8.0?

njaffer:
@mjgiarlo: Yes

jcoyne:
Here's the backtrace:
```
ActiveFedora::UnknownAttributeError - GenericWork does not have an attribute `ordered_member_ids':
  active-fedora (9.9.0) lib/active_fedora/attributes.rb:24:in `block in attributes='
  active-fedora (9.9.0) lib/active_fedora/attributes.rb:23:in `attributes='
  active-fedora (9.9.0) lib/active_fedora/attributes/serializers.rb:24:in `attributes='
  curation_concerns-models (0.8.0) app/actors/curation_concerns/base_actor.rb:62:in `apply_save_data_to_curation_concern'
  curation_concerns-models (0.8.0) app/actors/curation_concerns/base_actor.rb:24:in `create'
  curation_concerns-models (0.8.0) app/actors/curation_concerns/work_actor_behavior.rb:10:in `create'
  curation_concerns (0.8.0) app/controllers/concerns/curation_concerns/curation_concern_controller.rb:36:in `create'
```

jcoyne:
in #612 @tpendragon deletes the ordered_member_ids attribute in the update, but he's letting it pass through in the `create`, which is where @njaffer's problem arises.  I suspect the fix is to update the `WorkActorBehavior#create` to remove that attribute before it calls `super`.  Another possibility would be to update the Form object to not return that attribute when `WorkForm.model_attributes` is called.

jcoyne:
@njaffer I think your fix belongs here: https://github.com/projecthydra/sufia/blob/99a30f14b08c96dc9d4e9c0827620827ca071f0a/app/forms/sufia/forms/work_form.rb#L5  You should exclude :ordered_member_ids from the rendered terms on the form.

njaffer:
@jcoyne: ordered_members_ids are not rendered terms, so I cannot exclude it here.

PR #630, opened 2016-02-22
Allow create_content to take a relationship arg

jcoyne:
This enables use of the FileSetActor for files besides the original
file.  In our case we're using it for a preservation quality derivative
of the original.


escowles:
:+1:

PR #632, opened 2016-02-22
Write files to the staging area with their name

### LINKED ISSUE #631: When ingesting a second file, it uses the same staging name.
Calling `FileSetActor.create_content` with a relation on a second file causes it to have the name of the first file. This is because the FileSetActor was using the file_set.label (set only once) as the name for all attached files.

jcoyne:
Previously it was using the name of the first file for all subsequent
files. Fixes #631

Additionally this extracts FileActor from FileSetActor

### DIFF
--- a/spec/actors/curation_concerns/file_actor_spec.rb
+++ b/spec/actors/curation_concerns/file_actor_spec.rb
@@ -1,0 +1,18 @@
+require 'spec_helper'
+
+describe CurationConcerns::FileActor do
+  include ActionDispatch::TestProcess
+  let(:user) { create(:user) }
+  let(:file_set) { create(:file_set) }
+  let(:relation) { create(:file_set) }
+  let(:actor) { described_class.new(file_set, 'remastered', user) }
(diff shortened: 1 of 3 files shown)


jcoyne:
Build failed because it had a problem downloading solr. I was unable to restart it: https://github.com/travis-ci/travis-ci/issues/5678.  I think it's okay because the travis-ci/pr build worked.

awead:
:+1: after an initial read-through.

spec/actors/curation_concerns/file_actor_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'spec_helper'
+
+describe CurationConcerns::FileActor do
```
barmintor: TODO: Open a ticket to test that FileActor.revert_file dispatches to File#restore_version and CurationConcerns::VersioningService.create as appropriate to File#save

jcoyne: Those cases are still covered by tests in file_set_actor_spec.rb

PR #640, opened 2016-02-25
Make the views rely on local variables instead of ivars

jcoyne:


escowles:
What's the motivation for this change?  I see it simplifies the test a bit, but is there some other reason?

awead:
:+1: after rebase

awead:
@escowles it's kind of a preference, but using local variables allows you to override the view more easily.

jcoyne:
@escowles another reason for this change was to align the partials so they are all using the same local variable: https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/curation_concerns/base/show.html.erb#L9-L13

escowles:
@awead @jcoyne Thanks, those both sound like good things to make it easier for downstream customization of the views.  So I'm also :+1: after rebasing.

PR #643, opened 2016-02-25
Change name of local variable to presenter for consistency

jcoyne:
with other partials

### DIFF
--- a/app/views/curation_concerns/base/_representative_media.html.erb
+++ b/app/views/curation_concerns/base/_representative_media.html.erb
@@ -1,4 +1,4 @@
-<% if work.representative_id.present? && (fs = ::FileSet.load_instance_from_solr(work.representative_id)) %>
+<% if presenter.representative_id.present? && (fs = ::FileSet.load_instance_from_solr(presenter.representative_id)) %>
   <%= media_display fs %>
 <% else %>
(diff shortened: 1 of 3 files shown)


app/views/curation_concerns/base/_representative_media.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if work.representative_id.present? && (fs = ::FileSet.load_instance_from_solr(work.representative_id)) %>
+<% if presenter.representative_id.present? && (fs = ::FileSet.load_instance_from_solr(presenter.representative_id)) %>
```
awead: I know this isn't germane to the PR, but it seems like the presenter ought to be responsible for returning a FileSet, instead of putting this into the view, but if that's your usual MO for this, it's fine.

awead: For example, call `presenter.representative` and have it return the FileSet, loaded from solr, or nil.

jcoyne: Yes, good idea, but I don't think that's what *this* PR is really about. Can you make another ticket?

awead: @jcoyne sure!

PR #650, opened 2016-02-26
Move to DC Elements over DC Terms

tpendragon:
Curation Concerns has no built in way to handle URIs as values, so it
seems a better default would be predicates which don't enforce URIs via
their ranges.

### DIFF
--- a/curation_concerns-models/app/models/concerns/curation_concerns/basic_metadata.rb
+++ b/curation_concerns-models/app/models/concerns/curation_concerns/basic_metadata.rb
@@ -15,33 +15,33 @@ module CurationConcerns
       property :resource_type, predicate: ::RDF::Vocab::DC.type do |index|
         index.as :stored_searchable, :facetable
       end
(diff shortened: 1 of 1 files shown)


jcoyne:
:+1: but we should hear from others. The good thing about CC is that this is generated onto the model, so one could create a legacy metadata mixin if they don't want to move from DC terms

jcoyne:
Ping @cam156 @awead @njaffer @grosscol @mjgiarlo

awead:
fyi: http://wiki.dublincore.org/index.php/FAQ/DC_and_DCTERMS_Namespaces

:+1: Given it's a one to one mapping between the two namespaces but we aren't enforcing ranges are we? This isn't a problem, I'm just curious. At Penn state we can migrate predicates when we move to PCDM.

tpendragon:
> but we aren't enforcing ranges are we?

We kind of do the opposite of enforce - we provide predicates and only allow them to input invalid (according to the semantics of the predicate) data.

mcritchlow:
less directly involved here, atm, but big :+1: on this change

PR #651, opened 2016-02-29
Allow passing options to the resource routes

jcoyne:
This allow us to pass constraint options to the resource route methods
via the newly created `curation_concerns_named_resouce_route` method.

When upgrading, one will have to add `curation_concerns_resource_routes`
into `config/routes.rb`, as that's no longer part of
`curation_concerns_basic_routes`


jcoyne:
Hold off on merging this, I'm not sure it's helpful

PR #658, opened 2016-03-03
Return zero when members is null

awead:
refs projecthydra/sufia#1620

`member_ids_ssim` is sometimes not present in the solr document, yielding a NilClass error when calling `CurationConcerns::CollectionsPresenter.total_items`. This returns zero if the solr field is not present.


escowles:
:+1:

PR #659, opened 2016-03-04
Don't parse date_created as a Date, fixes #657

### LINKED ISSUE #657: Should date_created be indexed as a string instead of a date?
Note this comment in Sufia:

https://github.com/projecthydra/sufia/blob/master/app/presenters/sufia/collection_presenter.rb#L3

However to follow this recommendation would mean changing the index behavior. Presently, `date_created` is being indexed as a date. Do we want to change this and index it as a string?

awead:
Fixes #657 

`date_created` is indexed as a string, so we should display it like one.

* display date_created as string instead of trying to parse it like a Date


awead:
@jcoyne rebased.

PR #665, opened 2016-03-10
Updating local coverage handling

escowles:
Update coverage setup to work with local test run.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,4 +1,4 @@
-if ENV['CI']
+if ENV['CI'] || ENV['COVERAGE']
   require 'coveralls'
   Coveralls.wear!
 end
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -29,7 +29,7 @@
   require 'simplecov'
 
   SimpleCov.root(File.expand_path('../..', __FILE__))
-  SimpleCov.formatter = Coveralls::SimpleCov::Formatter
+  SimpleCov.formatters = [Coveralls::SimpleCov::Formatter, SimpleCov::Formatter::HTMLFormatter]
```
jcoyne: do we need both? Can it be SimpleCov `if ENV['CI'] ? Coveralls : HTML`?

escowles: Sure, we can do the machine-readable output for Travis, and HTML output otherwise.

PR #670, opened 2016-03-11
Cleaning before each example is slow, stop doing this.

jcoyne:


tpendragon:
According to https://travis-ci.org/projecthydra-labs/curation_concerns/pull_requests it doesn't seem like this gained us much

jcoyne:
It is true that we aren't seeing much in the way of speed improvements (most of the time on travis is spent running generators), however, this exercise did surface a few bugs.  I wonder if it isn't better this way, where you are explicit about each test that requires a clean repo.

tpendragon:
It seems strange to have to explicitly clean out your environment per test.

tpendragon:
I'm pretty :-1: on this one, can someone else from @projecthydra/sufia-code-reviewers comment?

mjgiarlo:
I'm +0 on this one. It'd be good for @tpendragon to be sold on it (and for this to be rebased).

PR #672, opened 2016-03-14
Convert Bytes to a stored long

tpendragon:
Solr has a 32-bit max value for integers, so collections with large files were quickly going over it and failing to index. By using a long instead it at least makes the number significantly larger before there's a problem.


jcoyne:
Looks, good, but wouldn't it be better to add `:long` here? https://github.com/projecthydra/solrizer/blob/05e6a01b32553b9df434eacc35c0888f8bbc8050/lib/solrizer/suffix.rb#L54-L68

tpendragon:
Probably, to be honest I was hesitant to add a PR to something that hasn't needed one in two years. ;) I'll send one.

PR #680, opened 2016-03-17
Use `feature` instead of `describe` for feature specs

jcoyne:

This allows us to remove an exception from .rubocop.yml
I've also reenabled the require of rubocop-rspec and pinned rubocop to
0.37 so we can still use it.


tpendragon:
Rebase

PR #687, opened 2016-03-17
WorkShowPresenter#file_presenters only returns FileSets

jcoyne:

This prevents members that are not FileSets from drawing in the section
of the page that is the list of attached files. For example if you have
a Book that has members that are Sections and FileSets we want to
exclude the Sections from displaying in the table of files.

Ref https://github.com/curationexperts/heliotrope/issues/84

### DIFF
--- a/app/presenters/curation_concerns/work_show_presenter.rb
+++ b/app/presenters/curation_concerns/work_show_presenter.rb
@@ -24,18 +24,41 @@ module CurationConcerns
              :creator, :contributor, :subject, :publisher, :language, :embargo_release_date,
              :lease_expiration_date, :rights, to: :solr_document
 
+    # @return [Array<FileSetPresenter>] presenters for the orderd_members that are FileSets
+    def file_set_presenters
+      @file_set_presenters ||= member_presenters(ordered_ids & file_set_ids)
+    end
+
+    # @deprecated
+    # @return [Array<FileSetPresenter>] presenters for the orderd_members that are FileSets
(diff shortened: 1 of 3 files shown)


jcoyne:
Final approval of this ticket should be given to @escowles because this may mess up how Plum works.

tpendragon:
Totally breaks plum, and doesn't make sense to me to treat filesets as special.

escowles:
@tpendragon What about renaming `file_presenters` to `member_presenters` and using the new behavior for `file_presenters`?  That keeps the full member list that Plum needs, but also provides an easy way to get just the filesets.

tpendragon:
@escowles Sure, but file_set_presenters.

jcoyne:
@tpendragon I will make this happen.

jcoyne:
@tpendragon do you like this version any better?

tpendragon:
Yeah, I like this better. The only issue is that there's nowhere which renders child works in our default UI now, whereas before it did, right? Of course, there's also nowhere to MAKE child works..

jcoyne:
@tpendragon right, you'll have to add a partial to show them. We do that like so:
https://github.com/curationexperts/heliotrope/blob/master/app/views/curation_concerns/monographs/show.html.erb#L11
You could also override https://github.com/projecthydra-labs/curation_concerns/blob/master/app/views/curation_concerns/base/_related_files.html.erb#L1 to use `member_presenters` instead of `file_set_presenters`

PR #690, opened 2016-03-18
WIP: Adding the browse_everything gem

awead:
This is the current state of adding the BrowseEverything gem into CurationConcerns. It's only working about halfway, but I have questions before it can be finished.

Here's how it presently looks:
cc-be

The upload UI has been ported over directly from Sufia, which means it includes the default upload from computer behavior. Is this what you want? Otherwise, we can remove that and have only the BE upload process?

This work was started at the dev meeting in San Diego last month. This is an effort to document how far we got. I may not have time to finish this, given Sufia 7 sprint requirements, so I'm marking where things are.

### DIFF
--- a/app/assets/javascripts/curation_concerns/application.js
+++ b/app/assets/javascripts/curation_concerns/application.js
@@ -13,3 +13,9 @@
 //= require jquery-ui/sortable
 //
 //= require curation_concerns/curation_concerns
+
+//
+// For Browse-everything until https://github.com/projecthydra-labs/browse-everything/issues/85 is resolved:
+//= require jquery.treetable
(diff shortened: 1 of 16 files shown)


jcoyne:
> Is this what you want?

Who is the "you" you are addressing?

tpendragon:
@awead :+1: to leaving the "upload from computer."

mjgiarlo:
@awead You planning to keep working on this PR or is it :skull:?

awead:
@mjgiarlo I don't have time at the moment, but I've left it open in case someone wants to pick it up.

mjgiarlo:
@awead okie dokie.

PR #696, opened 2016-03-28
Avoid hydra search builder

jcoyne:


lib/generators/curation_concerns/templates/catalog_controller.rb
```diff
@@ -1,7 +1,7 @@
 class CatalogController < ApplicationController
   include CurationConcerns::CatalogController
   configure_blacklight do |config|
-    config.search_builder_class = CurationConcerns::SearchBuilder
+    # config.search_builder_class = ::SearchBuilder
```
mjgiarlo: Do we want to leave this comment in?

jcoyne: Yes, that mirrors what is provided in the default Hydra CatalogController template

PR #702, opened 2016-03-30
Works and FileSets should support to_global_id

jcoyne:
So that you can send them to ActiveJob without writing a custom
serializer/deserializer.


mjgiarlo:
Well done! :clap:

PR #707, opened 2016-03-31
Refactor the actors into composable classes with a common interface

jcoyne:


app/services/curation_concerns/curation_concern.rb
```diff
@@ -1,9 +1,20 @@
 module CurationConcerns
   module CurationConcern
-    def self.actor(curation_concern, *args)
+    # Returns the top-level actor on the stack
+    def self.actor(curation_concern, current_user, attributes)
+      AddToCollectionActor.new(curation_concern, current_user, attributes,
```
tpendragon: Expectations on this API are a little odd. An Actor will take a subsequent actor AND call it? The subsequent actor can be either an array or a single instance (making the interface more than just sending a message.) I wonder if the solution here is to define a Pipeline with Pipes, and define an interface for a Pipe. Then the Pipeline's responsible for pushing everything through.

tpendragon:
:+1: on having a clear interface here. I'm not sure I agree with the actor's interface yet, but I like the direction it's headed.

app/services/curation_concerns/curation_concern.rb
jcoyne: This is actually a stack, forgive me for calling it a pipeline.  In the first pass ("down" if you will) in `initialize`, each node may strip off some of the attributes, and on the way up ("create/update") modifies the curation_concern.  I'm not sure how we can do that in a true pipeline.

mjgiarlo: Hereafter known as "Gobstoppers within gobstoppers" architecture.

tpendragon: ^ Right, this is the concerning bit for me. This might be my functional programming side talking, but actors should probably be units with a common API, which another pattern can take advantage of to move between them.

jcoyne: If we want this to be more functional we'd just need to have create and save create and return a state object that encapsulates status & parameters.  Would that be more to your liking?

jcoyne:
@tpendragon how do you like this new style?

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -115,7 +115,7 @@ def build_form
     end
 
     def actor
-      @actor ||= CurationConcerns::CurationConcern.actor(curation_concern, current_user, attributes_for_actor)
+      @actor ||= CurationConcerns::CurationConcern.actor(curation_concern, current_user)
```
jeremyf: I like this disentanglement. The `CurationConcerns::CurationConcern.actor` method is a query method (building an instance). The eventual `#update` method is the state changing thing based on attributes. It is the command.

jeremyf:
I like how the interface is shifting. Each actor has clear responsibility.

tpendragon:
@jcoyne @jeremyf Looks like I was a bit slow on the draw, I apologize. I have some comments about this code (it's better than the last time, but I wonder if next_actor is really necessary, and maybe the ActorStack should just iterate through its list of actors, making a nice Composite object). When there are cases like this where it's useful to merge iterative improvements, how do we talk about what the next iteration looks like?

jcoyne:
@tpendragon  It's a stack, not a list, because each actor on the stack can take action both before and/or after calling the next actor.

tpendragon:
@jcoyne Why is that? Is it important that keys be deleted off the attributes hash?

tpendragon:
(and then passed along, so it's like it's a new hash and not just mutated in memory)

jcoyne:
@tpendragon I'm not certain that that is important. I was just trying to preserve the existing behavior.

tpendragon:
In any case, this seems much more explicit than what we had in the past. It may be we'll want to look at separating some "reasons to change" out later, but we'll see.

Thanks for this, @jcoyne, it's awesome seeing the architecture improve!

PR #710, opened 2016-04-05
Reskin brand bar and title bar to use default bootstrap styling

cbeer:
Before (best case... there's lots of really bad looking edge cases too):
screen shot 2016-04-04 at 2 12 00 pm

After:
screen shot 2016-04-04 at 2 11 17 pm


mjgiarlo:
I'm :+1: on this. @escowles @tpendragon ?

escowles:
:+1: — much better to leave the styling as vanilla as possible to make it easier to override downstream.

mark-dce:
+1

PR #716, opened 2016-04-05
Merge curation_concerns-models into curation_concerns

cbeer:


mjgiarlo:
I am :+1: on this change, just like I was :+1: on merging sufia-models back into Sufia. No one is currently using cc-models, and I'm not aware of anyone planning to use cc-models independent of CurationConcerns.

I suspect we should wait for explicit :+1:s on this PR from the following people before we :shipit:

- [ ] @escowles 
- [ ] @tpendragon 
- [ ] @jpstroop

app/validators/has_one_title_validator.rb
```diff
@@ -0,0 +1,8 @@
+# validates that the title has at least one title
+class HasOneTitleValidator < ActiveModel::Validator
```
tpendragon: Where'd this come from?

config/locales/curation_concerns.en.yml
```diff
@@ -101,17 +102,29 @@ en:
         property: keywords
       title:
         property: name
+
```
tpendragon: Is this all new? I'm not saying I disagree with it or anything, but was it meant to be here?

lib/curation_concerns/engine.rb
```diff
@@ -18,6 +18,7 @@ class Engine < ::Rails::Engine
     require 'breadcrumbs_on_rails'
 
     config.autoload_paths += %W(
+      #{config.root}/app/actors/concerns
```
tpendragon: I'm surprised this is necessary. It breaks without it?

lib/curation_concerns/models.rb
```diff
@@ -0,0 +1,42 @@
+require 'hydra/head'
+
+module CurationConcerns
```
tpendragon: Where'd this stuff come from?

tpendragon: Ah, I see it.

tpendragon:
Couple comments inline, but I'm :+1: to the idea. cc-models always seemed questionable.

lib/curation_concerns/engine.rb
mjgiarlo: This matches my understanding of autoload, which is that Rails will handle `app/*/*.rb` and will handle the `concerns` subdirectory in directories it knows about (models, controllers -- not actors), but that it won't recursively load everything in `app/`.

config/locales/curation_concerns.en.yml
mjgiarlo: It's from https://github.com/projecthydra-labs/curation_concerns/pull/716/files/212f7496bcf388262e5ff46219722043cd319f21#diff-a47c94c05ae8b7b630d2d8c1ee12ff3d right?

cbeer:
Nope, none of this is new. A handful of things got renamed (like the `curation_concerns:models:install` generator), but the rest is directly copied.

jcoyne:
I'm not so sure about this.  How about the Fedora-Migrate use case. They'd just want CurationConcerns::Models, right?  They *could* use all of CC, but most of it is just extra dependencies.

app/validators/has_one_title_validator.rb
mjgiarlo: https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/validators/has_one_title_validator.rb

lib/curation_concerns/engine.rb
cbeer: This was directly from models. I'm :+1: on changing it, but not as part of this commit.

cbeer:
If the extra dependencies are not actively harmful, I suggest we make our own lives easier.

mjgiarlo:
Does Fedora-Migrate have a near-term roadmap that includes migrating F3 content directly to PCDM? Tagging @barmintor and @awead to sound off on this.

jpstroop:
I agree that this is primarily an artifact of sufia-models being separated from sufia, and I think the Gobstopper architecture generally solves that use case in a much better way, even if it's not perfect yet. I'd like @tpendragon and/or @escowles to confirm that this doesn't make writing custom concerns more difficult (I suspect the opposite), and that it's headed in the direction of patterns we'd like to support. Barring them surfacing anything like that, I'm :ok_hand: .

app/validators/has_one_title_validator.rb
tpendragon: Isn't it at `app/has_one_title_validator.rb` too then? There's a renamed without any changes part of this commit.

cbeer:
@jcoyne Does fedora-migrate use curation_concerns-models at all? I'm not seeing it in https://github.com/projecthydra-labs/fedora-migrate/blob/master/fedora-migrate.gemspec (although it may affect my other goal to merge in hydra-collections..)

mjgiarlo:
It doesn't currently, @cbeer. 0 other codebases are using cc-models.

barmintor:
I've been building a set of FedoraMigrate::Works mover impls, but I don't
think you'd package such a thing with Fedora Migrate.

app/validators/has_one_title_validator.rb
tpendragon: Yeah. @cbeer Can you get rid of https://github.com/projecthydra-labs/curation_concerns/blob/curation_concerns_models/app/has_one_title_validator.rb ?

cbeer: Directly from models. It's a confusing validator, and I'm not sure why other validators (like `presence`) are insufficient.

config/locales/curation_concerns.en.yml
cbeer: Nope, it was in models, and at least some of it is necessary. It's confusing to me what parts in the stack are using it, but I figure investigating that is separate from trying to merge the projects.

app/validators/has_one_title_validator.rb
tpendragon: No..I mean, there's two. One in app/validators and one in app.

tpendragon: The other one's here: https://github.com/projecthydra-labs/curation_concerns/blob/curation_concerns_models/app/validators/has_one_title_validator.rb

config/locales/curation_concerns.en.yml
barmintor: Iirc that validator is in the WorkBehavior that is included in the Sufia
work behavior for title...

app/validators/has_one_title_validator.rb
mjgiarlo: @tpendragon is right. I believe this should be nuked: https://github.com/projecthydra-labs/curation_concerns/pull/716/files#diff-aba3d48be12193802ced467a01472193

The validator should live in `app/validators/`.

tpendragon:
Alright, I'm :+1: now.

mjgiarlo:
OK, we haven't heard from @escowles yet.

@jcoyne do you still have concerns re: fedora-migrate?

jcoyne:
no further concerns

mjgiarlo:
If you are :+1:, feel free to merge, @escowles.

mjgiarlo:
:clap:

PR #719, opened 2016-04-07
Merge hydra-collections functionality into curation_concerns

cbeer:


cbeer:
I've tried to split functional changes from mechanical merging, although it was imperfect. This also leaves significant future cleanup, but I'd rather we tackle that after this initial merge.

mjgiarlo:
:clap:

PR #721, opened 2016-04-08
Work around for duplicate predicate issue

hectorcorrea:
This is to address issue https://github.com/projecthydra/sufia/issues/1693 reported in Sufia but that also happens in CurationConcerns.

The culprit is that both BasicMetadata (in CurationConcerns) and Characterization (in HydraWorks) define properties that map to the same predicate. By forcing BasicMedatata to be included before Characterization the issue does not happen (because Characterization detects the conflict via the AlreadyThereStrategy.)

To see the problem load the FileSet class via the Rails console and notice the warning:
```
$ bundle exec rails console
Loading development environment (Rails 4.2.6)
irb(main):001:0> FileSet
Same predicate (http://purl.org/dc/elements/1.1/creator) used for properties creator and creator
Same predicate (http://purl.org/dc/elements/1.1/language) used for properties language and language
=> FileSet
irb(main):002:0> exit
```

On a codebase with the fix proposed in this PR the warning is not issued.
```
$ bundle exec rails console
Loading development environment (Rails 4.2.6)
irb(main):001:0> FileSet
=> FileSet
irb(main):002:0> exit
```

See also this PR https://github.com/projecthydra-labs/hydra-works/pull/288

### DIFF
--- a/app/models/concerns/curation_concerns/file_set_behavior.rb
+++ b/app/models/concerns/curation_concerns/file_set_behavior.rb
@@ -1,6 +1,12 @@
 module CurationConcerns
   module FileSetBehavior
     extend ActiveSupport::Concern
+    # BasicMetadata needs to be included before Characterization since
+    # both of them declare properties with the same predicate (dc:creator
+    # and dc:language.) Loading BasicMetadata first allows Characterization
+    # to detect the duplicate (via the AlreadyThereStrategy) and prevents
+    # the warning.
+    include CurationConcerns::BasicMetadata
     include Hydra::Works::FileSetBehavior
     include Hydra::Works::VirusCheck
     include Hydra::Works::Characterization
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@hectorcorrea just one failure.

hectorcorrea:
@mjgiarlo yeah....I suspect Characterization is not adding `file_author` because there was already another property (BasicMetadata's creator) linked to the `dc:creator` predicate. Not sure what to do at this point? Any thoughts @jcoyne ?

mjgiarlo:
I'm sensing a pairing opportunity for this work. @hectorcorrea can you ping the #software-developers slack channel and see if anyone (@jcoyne ?) might be available?

hectorcorrea:
This depends on https://github.com/projecthydra-labs/curation_concerns/pull/721

grosscol:
I think what you've done here no longer depends on projecthydra-labs/hydra-works#288.  The renaming in hydra-works will probably relieve the include ordering issue for these  property names.  However, it should be noted or made explicit that HydraWork::Characterization comes with schemas.

mjgiarlo:
Thoughts on this, @escowles @tpendragon @jcoyne ?

jcoyne:
@hectorcorrea with the changes to hydra-works, do we still need the changes to `app/models/concerns/curation_concerns/file_set_behavior.rb` ?

hectorcorrea:
@jcoyne yes we do.

PR #722, opened 2016-04-13
Create a factory for the ActorStack

jcoyne:

This enables Sufia to customize the ActorStack slightly.

Fixes projecthydra/sufia#1776

Will be used to complete work on:
Ref projecthydra/sufia#1699

### DIFF
--- a/app/actors/curation_concerns/actor_stack.rb
+++ b/app/actors/curation_concerns/actor_stack.rb
@@ -1,0 +1,27 @@
+module CurationConcerns
+  class ActorStack
+    attr_reader :curation_concern, :user, :first_actor_class, :more_actors
+    def initialize(curation_concern, user, more_actors)
+      @curation_concern = curation_concern
+      @user = user
+      @more_actors = more_actors
+      @first_actor_class = @more_actors.shift || RootActor
+    end
+
+    def inner_stack
+      ActorStack.new(curation_concern, user, more_actors)
+    end
+
+    def actor
+      first_actor_class.new(curation_concern, user, inner_stack)
+    end
+
+    def create(attributes)
+      actor.create(attributes.with_indifferent_access)
+    end
+
+    def update(attributes)
+      actor.update(attributes.with_indifferent_access)
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


app/actors/curation_concerns/actor_stack.rb
```diff
@@ -0,0 +1,27 @@
+module CurationConcerns
+  class ActorStack
+    attr_reader :curation_concern, :user, :first_actor_class, :more_actors
+    def initialize(curation_concern, user, more_actors)
+      @curation_concern = curation_concern
+      @user = user
+      @more_actors = more_actors
+      @first_actor_class = @more_actors.shift || RootActor
```
cbeer: What's the difference between `RootActor` here and `BaseActor` below?

jcoyne: BaseActor is the super class for all *WorkActors.  RootActor is the terminator of the stack, it's follows the null object pattern.

cbeer: Got it. I wonder if we can give those things better names some time.

jcoyne: I'm open to suggestions.

mjgiarlo: `NullActor`?

jcoyne: @mjgiarlo I considered that, but the RootActor returns true for all methods, so calling it the NullActor seems confusing.

jcoyne: RecursiveBaseCaseActor?

jcoyne: In any case, neither of these classes were created in this PR, so if you've got a good name please open a ticket, or better yet, a PR.

mjgiarlo: Yeah, I wouldn't hold up the PR over the name. :shipit:

cbeer: Agreed.

PR #732, opened 2016-04-18
Unlink file_set from work when file_set is deleted

hectorcorrea:
When a `FileSet` is deleted we need to make sure the `GenericWork` does not point to it anymore. This PR makes sure that neither `GenericWork.thumbnail` nor `GenericWork.representative` point to a `FileSet` after being deleted.

This issue was first reported in `Sufia` https://github.com/projecthydra/sufia/issues/1852 but also applies to `CurationConcerns`.

Do not merge yet, I still need to add a couple of tests  before this is ready.

### DIFF
--- a/spec/actors/curation_concerns/file_set_actor_spec.rb
+++ b/spec/actors/curation_concerns/file_set_actor_spec.rb
@@ -151,21 +151,26 @@ describe CurationConcerns::FileSetActor do
       expect { file_set.reload }.to raise_error ActiveFedora::ObjectNotFoundError
     end
 
-    context "representative of a work" do
+    context "representative and thumbnail of a work" do
       let!(:work) do
         work = create(:generic_work)
         # this is not part of a block on the create, since the work must be saved
         # before the representative can be assigned
         work.ordered_members << file_set
         work.representative = file_set
+        work.thumbnail = file_set
         work.save
         work
       end
 
-      it "removes representative and the proxy association" do
-        expect(work.reload.representative_id).to eq(file_set.id)
+      it "removes representative, thumbnail, and the proxy association" do
+        gw = GenericWork.load_instance_from_solr(work.id)
+        expect(gw.representative_id).to eq(file_set.id)
+        expect(gw.thumbnail_id).to eq(file_set.id)
         expect { actor.destroy }.to change { ActiveFedora::Aggregation::Proxy.count }.by(-1)
-        expect(work.reload.representative_id).to be_nil
+        gw = GenericWork.load_instance_from_solr(work.id)
+        expect(gw.representative_id).to be_nil
+        expect(gw.thumbnail_id).to be_nil
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/curation_concerns/file_sets_controller_behavior.rb
```diff
@@ -71,6 +71,14 @@ def show
...
     def destroy
       parent = @file_set.in_works.first
+      # When running the demo application this line is needed to
+      # make sure the work does not point to the file_set that
+      # is being deleted. HOWEVER, the RSpec test added in this PR
+      # works even without it.
+      #
+      # Any ideas?
+      #
+      unlink_from_work(@file_set, parent)
```
hectorcorrea: @jcoyne @mjgiarlo: Any ideas why does the test **pass** without the fix but the internal application blows up without it??

jcoyne: Maybe because the internal application is getting the value out of solr instead of out of fedora?

app/controllers/concerns/curation_concerns/file_sets_controller_behavior.rb
```diff
@@ -124,6 +125,13 @@ def update_metadata
 
     protected
 
+      def unlink_from_work(file_set, work)
```
jcoyne: Can you move this method into `actor.destroy`?

hectorcorrea: @jcoyne Done.

spec/actors/curation_concerns/file_set_actor_spec.rb
```diff
@@ -151,21 +151,26 @@
...
       end
 
-      it "removes representative and the proxy association" do
-        expect(work.reload.representative_id).to eq(file_set.id)
+      it "removes representative, thumbnail, and the proxy association" do
+        gw = GenericWork.load_instance_from_solr(work.id)
+        expect(gw.representative_id).to eq(file_set.id)
+        expect(gw.thumbnail_id).to eq(file_set.id)
         expect { actor.destroy }.to change { ActiveFedora::Aggregation::Proxy.count }.by(-1)
-        expect(work.reload.representative_id).to be_nil
```
hectorcorrea: @jcoyne This is what is tricky about this code. 

Before the changes in this PR accessing the `work` from Fedora after deleting the `file_set` gave the correct result (i.e. `gw.thumbnail = nil`) However, accessing the `work` from Solr gave the wrong results (i.e. `gw.thumbnail_id = "old file set id"`)

I am not sure if we should add a ticket to ActiveFedora::Aggregation or ActiveFedora::Aggregation::Proxy to handle this at a lower level of the stack. 

For now this PR takes care of the issue.

jcoyne: Yes, please open a ticket in ActiveFedora for this problem. When that's done let's merge this fix.

hectorcorrea: Ticket has been added https://github.com/projecthydra/active_fedora/issues/1064

PR #736, opened 2016-04-20
Ignore page parameter on the collection query

hectorcorrea:
Fixes issue https://github.com/projecthydra/sufia/issues/1835 (reported in Sufia) 

When rendering the Collection show page the Solr query that fetches the collection information was picking up the `page` parameter that was intended for the members of the collection. 

When the user wanted to see "page 2 of collection members" the query to fetch the collection information was returning no results because of the "page 2" parameter. 

This PR fixes that issue. We honor the "page X" parameter only to fetch the members of the collection.


jcoyne:
This looks like the right fix, but could you please submit a test that shows it.

PR #738, opened 2016-04-20
Remove CollectionsController spec

jcoyne:

Because this is already covered by
`spec/controllers/curation_concerns/collections_controller_spec.rb`


jcoyne:
Coverage fell on this.  Let me work some more on it.

PR #739, opened 2016-04-20
Remove button helpers. They only add confusing indirection

jcoyne:


hectorcorrea:
@jcoyne a screenshot might be helpful here to see how the form looks before and after

jcoyne:
@hectorcorrea there is no functional or visual change here.

hectorcorrea:
@jcoyne ah...my bad, I thought "UI helpers" instead of "Rails helpers" :ghost:

PR #746, opened 2016-04-25
Refactor test to expect a method call

jcoyne:

Instead of a generic error. The old way caused an rspec deprecation
warning.


mjgiarlo:
The descriptions of the changed specs say that they expect a redirect, but it doesn't seem you're testing that a redirect happens.

jcoyne:
The `authenticate_user!` causes a redirect to the devise sign-in controller when the user isn't signed in.

mjgiarlo:
I.e., CurationConcerns does not need to test devise. That seems reasonable to me.

PR #750, opened 2016-04-28
Display an error if the FileSet can't be updated

jcoyne:
Previously it was just silently failing.

Ref https://github.com/projecthydra/sufia/issues/1933


mjgiarlo:
Coveralls result looks bogus to me, so I'm :+1:

PR #753, opened 2016-04-28
Run callbacks for curation concern lifecycle events

### LINKED ISSUE #752: Run callbacks for concern lifecycle events
Callbacks are only firing for `FileSets` at the moment, and not for other concerns (e.g., `GenericWorks` in Sufia). Callbacks should be fired when any curation concern is created, updated, or destroyed.

Blocks https://github.com/projecthydra/sufia/issues/1437

mjgiarlo:
Split `:after_create_content` callback into `:after_create_concern` and `:after_create_fileset` since we're already special-casing FileSets; this allows us to differentiate between attaching FileSets and creating other concerns.

Fixes #752

### DIFF
--- a/app/actors/curation_concerns/base_actor.rb
+++ b/app/actors/curation_concerns/base_actor.rb
@@ -11,16 +11,21 @@ module CurationConcerns
       @cloud_resources = attributes.delete(:cloud_resources.to_s)
       apply_creation_data_to_curation_concern
       apply_save_data_to_curation_concern(attributes)
-      next_actor.create(attributes) && save
+      next_actor.create(attributes) && save && run_callbacks(:after_create_concern)
     end
 
     def update(attributes)
       apply_update_data_to_curation_concern
       apply_save_data_to_curation_concern(attributes)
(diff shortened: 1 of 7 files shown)


app/actors/curation_concerns/base_actor.rb
```diff
@@ -11,17 +11,22 @@ def create(attributes)
       @cloud_resources = attributes.delete(:cloud_resources.to_s)
       apply_creation_data_to_curation_concern
       apply_save_data_to_curation_concern(attributes)
-      next_actor.create(attributes) && save
+      next_actor.create(attributes) && save && run_callbacks(:after_create_content)
```
jcoyne: This isn't really "create_content".  That was for a binary deposit. This is "create metadata". Can we make a new callback type?

mjgiarlo: In this context, I interpreted "content" as the Work itself.

jcoyne: I think we should keep those distinct. I think we ought to create a "create_work" callback.

PR #758, opened 2016-04-29
Remove Resque from the gem

jcoyne:

This enables applications to choose their own queuing backend such as Sidekiq.


mjgiarlo:
@jcoyne so... now we should:

- [ ] Cut a new release of CC now/soon
- [ ] Make the necessary changes in Sufia to get it working atop the release (stripping out all the resque vestiges)

:question:

PR #760, opened 2016-05-02
Remove poorly named methods

jcoyne:

GenericWork is not part of the model. Let's not have methods that imply
otherwise.


mjgiarlo:
Subset of #761. Will incorporate this into that PR.

PR #761, opened 2016-05-02
Nuke outdated references to GenericWorks and GenericFiles

mjgiarlo:
Paraphrasing #760's rationale: GenericWork is not part of CurationConcerns models. Let's not have methods that imply otherwise.

Note that there are still a bunch of files under `spec/` that continue to reference GWs, but that's because `GenericWork` is the CC work generated by the test app generator. I don't care about those.

Closes #760.

### DIFF
--- a/app/models/concerns/curation_concerns/file_set/belongs_to_works.rb
+++ b/app/models/concerns/curation_concerns/file_set/belongs_to_works.rb
@@ -7,19 +7,14 @@ module CurationConcerns
         before_destroy :remove_representative_relationship
       end
 
-      def generic_works
-        in_objects # in_objects is provided by Hydra::PCDM::ObjectBehavior
-      end
-
-      # OPTIMIZE: We can load this from Solr much faster than loading the objects
-      def generic_work_ids
-        generic_works.map(&:id)
+      def parents
+        in_works
       end
 
       # Returns the first parent object
       # This is a hack to handle things like FileSets inheriting access controls from their parent.  (see CurationConcerns::ParentContainer in app/controllers/concerns/curation_concers/parent_container.rb)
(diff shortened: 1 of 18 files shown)


app/models/concerns/curation_concerns/file_set/belongs_to_works.rb
```diff
@@ -7,41 +7,41 @@ module BelongsToWorks
...
 
       # Files with sibling relationships
-      # Returns all FileSets aggregated by any of the GenericWorks that aggregate the current object
+      # Returns all FileSets aggregated by any of the parent objects that
+      # aggregate the current object
       def related_files
-        generic_works = self.generic_works
-        return [] if generic_works.empty?
-        generic_works.flat_map { |work| work.file_sets.select { |file_set| file_set.id != id } }
+        return [] if parents.empty?
```
jcoyne: This calls `parents` twice, which triggers 2x the loads from Fedora, right?

app/models/concerns/curation_concerns/file_set/belongs_to_works.rb
```diff
@@ -7,41 +7,41 @@ module BelongsToWorks
...
       end
 
-      # If any parent works are pointing at this object as their representative, remove that pointer.
+      # If any parent objects are pointing at this object as their
+      # representative, remove that pointer.
       def remove_representative_relationship
-        generic_works = self.generic_works
-        return if generic_works.empty?
-        generic_works.each do |work|
+        return if parents.empty?
```
jcoyne: Again here.

escowles:
Looks good to me -- great to clean up some of that confusing outdated terminology

PR #762, opened 2016-05-02
Don't call #parents twice. Expensive.

mjgiarlo:
I did this in #761. Oops. Thanks for finding this, @jcoyne


cbeer:
If it's so expensive, should it be memoized?

jcoyne:
@cbeer Yes, please do that 😉

PR #767, opened 2016-05-05
Method for updating rights and supporting retired rights.

### LINKED ISSUE #747: Allow marking which rights are currently active.
### Descriptive summary
Wanted to update the Rights available to the user from offering a selection of Creative Commons 3.0 to 4.0 licenses.  Simply adding licenses makes both the new and old available.  Removing the old licenses breaks the application due to unresolvable terms for the old ids (licenses). 

This came up using CurationConcerns 0.11.  The issue persists to HEAD.

The initial solution landed upon was to add an `active` key to each term.  See related work. 

### Expected behavior
Should be able to update licenses offered, mark the previous ones as no longer active, and support the previous licenses.

`RightsService.select_options` should only display the current options.

### Actual behavior
Adding licenses results in all being offered, and removing previous licenses breaks the QA `label` lookup.  There is no way to mark terms as active or inactive.

`RightsService.select_options` displays all options in rights.yml.

1. Create object with rights.
2. Remove or modify the rights id in `config/authorties/rights.yml`
3. View object

### Related work
Have already solved this issue with a monkey patch. mlibrary/umrdr#326

grosscol:
Fixes #747

Allow adding new rights and marking retired rights as inactive.  Still resolve labels for both.

This supports resolving the labels for rights while not offering them as current choices to the users.

* Add active attribute
* Make rights service utilize active attribute of terms.
* Add fixture file of dummy rights for testing.
* Update rights tests to use fixture file.

### DIFF
--- a/spec/services/rights_service_spec.rb
+++ b/spec/services/rights_service_spec.rb
@@ -1,17 +1,29 @@
 require 'spec_helper'
 
 describe RightsService do
-  describe "select_options" do
-    subject { described_class.select_options }
+  before do
+    # Configure QA to use fixtures
+    qa_fixtures = { local_path: File.expand_path('../../fixtures/authorities', __FILE__) }
+    stub_const("Qa::Authorities::LocalSubauthority::AUTHORITIES_CONFIG", qa_fixtures)
+  end
 
-    it "has a select list" do
-      expect(subject.first).to eq ["Attribution 3.0 United States", "http://creativecommons.org/licenses/by/3.0/us/"]
+  describe "#select_options" do
+    it "returns active terms" do
+      expect(described_class.select_options).to include(["First Active Term", "demo_id_01"], ["Second Active Term", "demo_id_02"])
+    end
+
+    it "does not return inactive terms" do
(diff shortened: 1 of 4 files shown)


jcoyne:
Looks like rubocop was unhappy.

spec/services/rights_service_spec.rb
```diff
@@ -1,17 +1,29 @@
 require 'spec_helper'
 
 describe RightsService do
-  describe "select_options" do
-    subject { described_class.select_options }
+  before do
+    # Configure QA to use fixtures
+    qa_fixtures = { local_path: File.expand_path('../../fixtures/authorities', __FILE__) }
+    stub_const("Qa::Authorities::LocalSubauthority::AUTHORITIES_CONFIG", qa_fixtures)
```
grosscol: Had to override this constant in order to get QA to read from a different authority file.  Is this an okay way to go about doing this?

jcoyne: We'll it's not optimal that QA doesn't have a better way to do this, but not your problem. 👍

Issue #766, opened 2016-05-05
Improve documentation of the concepts introduced in /app


jeremyf:
For those that have completed their Rails 101 experience, the concepts of Helper, Models, Controllers, and Views are well explained. However, CurationConcerns adds `Actors`, `Forms`, `Indexers`, `Inputs`, `Jobs`, `Presenters`, `Renderers`, `SearchBuilders`, `Services`, `Validators`.

For reference, [Sipity](//github.com//ndlib/sipity) has documentation of some of the concepts that can be visited and provide some guidance.
- [x] Actors
- [x] Forms
- [ ] Indexers
- [ ] Inputs
- [ ] Jobs
- [ ] Presenters
- [ ] Renderers
- [ ] SearchBuilders
- [ ] Services
- [ ] Validators

escowles:
There is a Vagrant project to spin up a stock CurationConcerns app, with some accompanying instructions for customizing that covers some of this:

https://github.com/escowles/curation-concerns-vagrant/wiki/Dive-Into-Curation-Concerns

mjgiarlo:
Unassigning this and moving back to the backlog.

PR #768, opened 2016-05-05
add form documentation

carrickr:
refs #766 

Improvement for documentation of Forms and a future forms model to group al forms in.

@elrayle @jcoyne This is based off @jeremyf 's explanation of forms, but we may need to look deeper into this regarding validation, its model wrapping, etc.

### DIFF
--- a/app/forms/curation_concerns/forms.rb
+++ b/app/forms/curation_concerns/forms.rb
@@ -1,14 +1,13 @@
 module CurationConcerns
-	 # This will be the module for containing forms
-	 # @since 0.14.0
-	# Forms are a type of data structures that include validation that leverage hydra editor 
-	# capabilities. Their purpose is to allow users to edit and save objects based on models.
-	# The impetus for their creation was to remove validation from models, and perhaps 
-	# demultiplex fields from one form to multiple objects. A form must define the model
-	# it uses and attributes it manipulates. It may include default values and define required
-	# required fields. They are ruby objects that are typically instantiated by the controller 
-	# to facilitate user creation, editing, or deleting of an object. They can also be used to
-	# sanitize attributes passed along to actors, as they are in curation concerns.
-	module Forms
-	end
+ # This will be the module for containing forms
+ # @since 0.14.0
+ # Forms are a type of data structures that include validation that leverage hydra editor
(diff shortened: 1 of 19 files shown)


app/forms/curation_concerns/forms.rb
```diff
@@ -0,0 +1,13 @@
+module CurationConcerns
+ # This will be the module for containing forms
+ # @since 0.14.0
+ # Forms are a type of data structures that include validation that leverage hydra editor
+ # capabilities. Their purpose is to allow users to edit and save objects based on models.
+ # The impetus for their creation was to remove validation from models. A form must define
+ # the model it uses and attributes it manipulates. It may include default values and define
+ # required fields.
+ # They are ruby objects that are typically instantiated by the controller to facilitate
+ # user creation, editing, or deleting of an object.
```
jeremyf: I believe my statement about forms reflected my understanding from Sipity (in part inspired by https://github.com/apotonick/reform) in which validations are performed. As I'm reviewing forms in Curation Concerns it appears they are used to sanitize attributes passed along to actors.

I don't know if the "The impetus for their creation was to remove validation from models." holds true within the context of CurationConcerns.

Perhaps @jcoyne or @dchandekstark can weight in?

jcoyne: Remove validations from models, and perhaps demultiplex fields from one form to multiple objects.

gordonleacock: module CurationConcerns
	 # This will be the module for containing forms
	 # @since 0.14.0
	# Forms are a type of data structures that include validation that leverage hydra editor 
	# capabilities. Their purpose is to allow users to edit and save objects based on models.
	# The impetus for their creation was to remove validation from models, and perhaps 
	# demultiplex fields from one form to multiple objects. A form must define the model
	# it uses and attributes it manipulates. It may include default values and define required
	# required fields. They are ruby objects that are typically instantiated by the controller 
	# to facilitate user creation, editing, or deleting of an object. They can also be used to
	# sanitize attributes passed along to actors, as they are in curation concerns.
	module Forms
	end
end

jcoyne:
It looks like this branch needs to be rebased against master.

PR #773, opened 2016-05-10
Dump unused Redis config value.

mjgiarlo:
The `config.redis_namespace` value is largely vestigial in CC at this point; it is used only by a migrate task that turns `ChecksumAuditLog` entries about `GenericFiles` into ones about `FileSets`. CC has no legacy data with `GenericFiles`, unless we are supporting a Curate- or Worthwhile-to-CC path, for which I am seeing a lack of tooling and documentation. The migrate task I've deleted in this PR [still lives in Sufia](https://github.com/projecthydra/sufia/blob/master/tasks/migrate.rake#L14), where there's more tooling for upgrades.

This PR provides a cleaner boundary between Sufia, which cares about Redis for activity streams where namespacing is relevant, and CC, which uses Redis only as a locking mechanism.


grosscol:
Shouldn't curation_concerns/lib/generators/curation_concerns/templates/config/redis_config.rb be dumped as well?

grosscol:
The spec for the task you removed fails.  Remove the spec for migrate.rake as well.

mjgiarlo:
@grosscol just pushed that. thx!

lib/tasks/migrate.rake
```diff
@@ -1,11 +0,0 @@
-namespace :curation_concerns do
-  namespace :migrate do
-    desc "Migrate audit logs"
-    task audit_logs: :environment do
```
jcoyne: Is this required for a sufia 6 -> 7 migration?

jcoyne: Nope, it's here: https://github.com/projecthydra/sufia/blob/master/tasks/migrate.rake#L14

mjgiarlo:
@grosscol :speech_balloon: 
> Shouldn't curation_concerns/lib/generators/curation_concerns/templates/config/redis_config.rb be dumped as well?

I'm not sure. Does CC's usage of Redis for locking make use of or require this config? Tagging @tpendragon @escowles @jcoyne

jcoyne:
Redis is required.  I *Think* we could dump the part about Passenger from it, but we still need the config.

tpendragon:
I'm not sure. redis' config seems like it's a singleton at times too.

mjgiarlo:
I've included a simplification of `redis_config.rb` in the PR now that:

1. Declares CC doesn't care about Passenger (it's an application-level concern).
2. Breaks when a Redis connection cannot be made, rather than setting `Redis.current` to `nil` and shrugging.

 Questions/concerns about that are welcome.

PR #776, opened 2016-05-10
Use hydra-head 10

jcoyne:


jcoyne:
This patch to hydra-head should fix the remaining error: https://github.com/projecthydra/hydra-head/pull/350

PR #783, opened 2016-05-12
Characterization metadata now stored in original_file

awead:
Relates to projecthydra/sufia#1894

Uses changes in HydraWorks to keep technical metadata on the original_file, but accessible through a proxy mechanism defined in `CurationConcerns::Characterization`

* technical metadata is delegated to `original_file`
* `FileSet.find` must be used in place of `FileSet.load_instance_from_solr` in certain cases

### DIFF
--- a/app/models/concerns/curation_concerns/characterization.rb
+++ b/app/models/concerns/curation_concerns/characterization.rb
@@ -1,0 +1,41 @@
+# This module points the FileSet to the location of the technical metdata.
+# By default, the file holding the metadata is :original_file and the terms
+# are listed under ::characterization_terms.
+# Implementations may define their own terms or use a different source file, but
+# any terms must be set on the ::characterization_proxy by the Hydra::Works::CharacterizationService
+#
+# class MyFileSet
+#   include CurationConcerns::FileSetBehavior
+# end
+#
+# MyFileSet.characterization_proxy = :master_file
+# MyFileSet.characterization_terms = [:term1, :term2, :term3]
+module CurationConcerns
+  module Characterization
+    extend ActiveSupport::Concern
+
+    included do
+      class_attribute :characterization_terms, :characterization_proxy
+      self.characterization_terms = [
+        :format_label, :file_size, :height, :width, :filename, :well_formed,
+        :page_count, :file_title, :last_modified, :original_checksum, :mime_type
+      ]
+      self.characterization_proxy = :original_file
+
+      delegate(*characterization_terms, to: :characterization_proxy)
+
+      def characterization_proxy
+        send(self.class.characterization_proxy) || NullCharacterizationProxy.new
+      end
+    end
+
+    class NullCharacterizationProxy
+      def method_missing(*_args)
+        []
+      end
+
+      def mime_type
+      end
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


app/views/curation_concerns/base/_representative_media.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if presenter.representative_id.present? && (fs = ::FileSet.load_instance_from_solr(presenter.representative_id)) %>
+<% if presenter.representative_id.present? && (fs = ::FileSet.find(presenter.representative_id)) %>
```
jcoyne: we need to get rid of this find. Preferably by putting it into the presenter.

awead: yes. yes, we do. Do you want to do that in this PR?

mjgiarlo: Waiting on a resolution to this thread before merging.

mjgiarlo:
Needs a squash too.

app/views/curation_concerns/base/_representative_media.html.erb
jcoyne: Either resolve it now, or ticket it.

awead: @jcoyne @mjgiarlo I'm inclined to ticket this and mark it for 7.0.0. We could then get a beta out this week.

spec/factories/file_sets.rb
```diff
@@ -5,6 +5,7 @@
     transient do
       user { FactoryGirl.create(:user) }
       content nil
+      filename nil
```
jcoyne: Why do we need this?

app/views/curation_concerns/base/_representative_media.html.erb
mjgiarlo: :ok_hand: Can you drop a ticket # in here, and then I'll merge when the build is :green_apple:

awead: @mjgiarlo @jcoyne https://github.com/projecthydra-labs/curation_concerns/issues/785

app/models/concerns/curation_concerns/characterization.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      def self.characterization_proxy
+        :original_file
+      end
+
+      delegate(*characterization_terms, to: :characterization_proxy)
+
+      def characterization_proxy
+        return NullCharacterizationProxy.new if send(self.class.characterization_proxy).nil?
```
mjgiarlo: Will these two lines result in loading the `:original_file` from Fedora twice in a row? If so, include in the follow-up ticket.

app/models/concerns/curation_concerns/characterization.rb
```diff
@@ -0,0 +1,47 @@
...
+        send(self.class.characterization_proxy)
+      end
+    end
+
+    class NullCharacterizationProxy
+      def method_missing(*_args)
```
mjgiarlo: Are you using this to shortcut having the null proxy respond to all the messages that map to the symbols in `self.characterization_terms`?

app/models/concerns/curation_concerns/characterization.rb
awead: @mjgiarlo good catch. I have a fix.

app/models/concerns/curation_concerns/characterization.rb
awead: Yes, with the exception of `mime_type`, which is a singular value returned from calling mime_type on Hydra::PCDM::File. Anything else would be an RDF property, which for AF::File and H::P::File, is a plain old AT::Resource and is always an array.

mjgiarlo: OK. I was briefly considering suggesting that you loop over `self.characterization_terms` and define those methods rather than use `method_missing` but meh, I think we should merge and move on.

awead:
🎉

PR #786, opened 2016-05-13
Make the ingest activejob queue name configurable

cbeer:
I'm on the fence whether this is worth the added complexity. For hybox, we're thinking about pushing all jobs into the same queue so we can manage and scale it easily.

I'm also tempted to make the default to just use the Rails default queue, and make using a separate queue an additional configuration.

### DIFF
--- a/lib/curation_concerns/configuration.rb
+++ b/lib/curation_concerns/configuration.rb
@@ -127,6 +127,13 @@ module CurationConcerns
       @lock_retry_delay ||= 200 # milliseconds
     end
 
+    # @!attribute [w] ingest_queue_name
+    #   ActiveJob queue to handle ingest-like jobs.
+    attr_writer :ingest_queue_name
+    def ingest_queue_name
+      @ingest_queue_name ||= :ingest
+    end
+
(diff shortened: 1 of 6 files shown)


cam156:
👍  Looks like the tests are not passing due to the release of a new HydraWorks which moved characterization.

lib/curation_concerns/configuration.rb
```diff
@@ -127,6 +127,13 @@ def lock_retry_delay
       @lock_retry_delay ||= 200 # milliseconds
     end
 
+    # @!attribute [w] ingest_queue_name
+    #   ActiveJob queue to handle ingest-like jobs.
+    attr_writer :ingest_queue_name
+    def ingest_queue_name
```
mjgiarlo: Should this also be exposed in `lib/generators/curation_concerns/templates/config/curation_concerns.rb` so folks know they can tweak it in their apps?

cbeer: Given the discussion #788, I'd almost want to leave it as a "beta" setting until we are all happy with a pattern.

PR #788, opened 2016-05-13
Put the ResolrizeJob on the default queue

cbeer:
In #784, we collapsed most of the job queues into `:ingest` or the default queue. We were uncertain what to do with the `ResolrizeJob` (which is only invoked via a Rake task). It seems like we have two choices here:

- a new queue, with a slightly more abstract name (names welcome..)
- the default queue, recognizing it could occupy a worker thread.

The default queue seems to be the least painful for new or small-scale adopters, although larger adopters may want to configure this in an initializer.

### DIFF
--- a/app/jobs/resolrize_job.rb
+++ b/app/jobs/resolrize_job.rb
@@ -1,6 +1,4 @@
 class ResolrizeJob < ActiveJob::Base
-  queue_as :resolrize
-
   def perform
     ActiveFedora::Base.reindex_everything
   end


cam156:
@cbeer Part of the reason we put the resolrize on its own queue is that we run it each time we deploy new code to make sure our index is correct.  It is a very long running job, so we wanted other things to run in parallel (things being characterization and batch update ate the time).  I'm not sure what the state of the jobs is at this point.  Maybe a configurable queue?

jcoyne:
@cam156 if you only kick off one copy of that job (via a rake task), how could it possibly be blocking anything else? The only way I see this happening is if you kickoff > n reindex jobs, where n is the number of workers.

tpendragon:
I wish I had seen this before #784. I'm not sure how i feel about this - at Princeton we've been really happy with the ability to reprioritize queues so easily. Pushing everything down into ingest means we can't focus down what's important.

For instance, it's FAR more important to us that the file's in Fedora than the derivatives/characterization/etc are run, because those are long running jobs that can take as long as we need them to. With #784, and this to a much lesser degree, we can't pick.

tpendragon:
I think what I'd be more comfortable with is three-four queues (critical?/high/medium/low), and then configuration options for the worker types so we can adjust where we want them to happen.

cbeer:
@tpendragon I think you should be able to customize the queues in an initializer, e.g.

```ruby
CharacterizeJob.queue_as :characterize
```

tpendragon:
@cbeer I could, but it spreads out configuration. Why not move #786 around a tiny bit to allow for `CurationConcerns.config.ingest_queue_name(self.class)` in the job, where it'll default down to medium?

atz:
This is the one job that I'm pretty sure should be kept separate from the others.  The duration of execution could easily be hours or even *days*.  Sure, it doesn't block if you have enough other workers, but.... in SQS you should obviously have different attributes associated with this queue.  In particular, the long execution will exceed the default `VisibilityTimeout`, meaning your other worker(s) will also pick it up and begin reindexing.  I can imagine *dozens* of workers all reindexing the same stuff in parallel.  So the availability of other workers is a logical and performance **detriment** if you can't keep `VisibilityTimeout` tuned to something reasonable for execution duration.

Clearly you cannot make `VisibilityTimeout` arbitrarily large just to accommodate this task, because if 99% of the other messages in this queue are timely in execution, when there is an inevitable hiccup, their retries are deferred unreasonably.  

I would revert.

cam156:
@atz That is another reason to keep the job segregated.  With resque you do not get different time outs so I have not configured that in the past.  

In practice we have only kept one default queue so that it executes only one item at a time.  If testing shows that there are no timing issues in Sufia 7 then having multiple default workers is great.  Unfortunately we never saw Fedora successfully lock for the timing issues we were seeing in Sufia, and I am not sure that it would ever.  If I load a File, run some long task and then update the metadata on the file I loaded a while back Fedora would not know that I have old metadata and overwrite any new metadata I have changed in the intervening time.

PR #793, opened 2016-05-17
WIP: Use the characterization proxy for the CharacterizeJob

### LINKED ISSUE #792: CharacterizeJob should pass the original_file
CharacterizeJob is passing the FileSet instead of the original file.

awead:
Fixes #792 

CharacterizationJob needs the file instead of the file set. Use the proxy that's defined as `characterization_proxy` on the file set, which by default is `:original_file`

### DIFF
--- a/app/jobs/characterize_job.rb
+++ b/app/jobs/characterize_job.rb
@@ -1,11 +1,28 @@
 class CharacterizeJob < ActiveJob::Base
   queue_as CurationConcerns.config.ingest_queue_name
 
+  before_perform :wait_for_content # do |job|
+  #  byebug
+  # end
+
+  DEFAULT_WAIT_TIME = 0.01
+  DEFAULT_WAIT_COUNT = 10
+
   # @param [FileSet] file_set
   # @param [String] filename a local path for the file to characterize. By using this, we don't have to pull a copy out of fedora.
+  # @param [Hash] opts hash of options for wait_for_content: :wait_time, :wait_count
   def perform(file_set, filename)
-    Hydra::Works::CharacterizationService.run(file_set, filename)
+    Hydra::Works::CharacterizationService.run(file_set.characterization_proxy, filename)
     file_set.save!
     CreateDerivativesJob.perform_later(file_set, filename)
+  end
+
+  def wait_for_content
+    count = 0
+    until arguments.first.original_file.content
+      sleep DEFAULT_WAIT_TIME
+      count += 1
+      raise(IOError, 'no content found for :original_file') if count == DEFAULT_WAIT_COUNT
+    end
   end
 end
--- a/spec/jobs/characterize_job_spec.rb
+++ b/spec/jobs/characterize_job_spec.rb
@@ -1,18 +1,32 @@
 require 'spec_helper'
 
 describe CharacterizeJob do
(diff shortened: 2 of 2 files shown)


app/jobs/characterize_job.rb
```diff
@@ -1,11 +1,28 @@
 class CharacterizeJob < ActiveJob::Base
   queue_as CurationConcerns.config.ingest_queue_name
 
+  before_perform :wait_for_content # do |job|
+  #  byebug
+  # end
+
+  DEFAULT_WAIT_TIME = 0.01
+  DEFAULT_WAIT_COUNT = 10
+
```
awead: should these be configured in CC initializer config?

mjgiarlo:
Hrmm. If ordering is now a concern, does this suggest that we should rejigger our code paths so that characterizejob doesn't kick off until files are attached? I'm leery of a sleeping/waiting approach.

awead:
@mjgiarlo how would you suggest doing that?

mjgiarlo:
@awead I can't say I've looked into it very deeply, but what if `CharacterizeJob` is called from e.g. `IngestFileJob` after files have been ingested instead of directly from within the actor?

awead:
@mjgiarlo that could work, but what about `CreateDerivativesJob`? Should we just lump all three of them together?

mjgiarlo:
@awead Meh. If one job calls another job that calls another job, I'm OK with that.

awead:
@mjgiarlo there's still may be the potential for race conditions:
```
class BigJob < ActiveJob::Base
  def perform(file_set)
    IngestFileJob.perform_later(file_set)
    CharacterizeJob.perform_later(file_set)
    CreateDerivativesJob.perform_later(file_set)
  end    
end
```
Unless they're all called via `.perform_now` I'm not sure there's any guarantee that they get done in that order.

mjgiarlo:
@awead Our code looks different from that now, though, right? Each job issues a save before invoking other jobs, no? The `BigJob` approach would indeed be problematic.

app/jobs/characterize_job.rb
jeremyf: Yes, if possible.

app/jobs/characterize_job.rb
```diff
@@ -1,11 +1,28 @@
...
   def perform(file_set, filename)
-    Hydra::Works::CharacterizationService.run(file_set, filename)
+    Hydra::Works::CharacterizationService.run(file_set.characterization_proxy, filename)
     file_set.save!
     CreateDerivativesJob.perform_later(file_set, filename)
   end
+
+  def wait_for_content
+    count = 0
+    until arguments.first.original_file.content
```
jeremyf: Are we reading content each time? Will that return the entire content binary?

cbeer:
We can also invoke one as "now" and the others as "later", e.g.: 

```ruby
class BigJob < ActiveJob::Base
  def perform(file_set)
    IngestFileJob.perform_now(file_set)
    CharacterizeJob.perform_later(file_set)
    CreateDerivativesJob.perform_later(file_set)
  end    
end
```

awead:
@cbeer yeah, although I think we're now looking at the kickoff approach. see #795

awead:
This needs to be reworked in light of #795

PR #794, opened 2016-05-17
Changing media display to use solr document instead of pulling the fille set from fedora

### LINKED ISSUE #785: Remove calls to ::FileSet.find
### Descriptive summary
Refactor this line https://github.com/projecthydra-labs/curation_concerns/blob/fded58d4c3edb2c1dad8cabc14236d18e874085e/app/views/curation_concerns/base/_representative_media.html.erb#L1 so that:

- [ ] The logic is encapsulated within the presenter
- [ ] It loads from Solr (remove `FileSet.find` which loads from Fedora)

### Expected behavior

Calling find on FileSet should load additional technical metadata properties from original_file

### Actual behavior

`original_file` is being loaded from Fedora (not solr).

cam156:
Fixes #785 

Using SolrDocument which includes the mime types and the representative ID to display the media instead on using the fedora files set.

* Use presenter and solr document instead of fedora file set

### DIFF
--- a/app/helpers/curation_concerns/file_set_helper.rb
+++ b/app/helpers/curation_concerns/file_set_helper.rb
@@ -7,9 +7,12 @@ module CurationConcerns::FileSetHelper
     end
   end
 
-  def media_display(file_set, locals = {})
-    render media_display_partial(file_set),
-           locals.merge(file_set: file_set)
+  # REVIEW: Since this media display could theoretically work for
+  #         any object that inplements to_s and the Mime Type methos (image? audio? ...),
+  #         Should this really be in file_set or could it be in it's own helper class like media_helper?
+  def media_display(presenter, locals = {})
+    render media_display_partial(presenter),
+           locals.merge(file_set: presenter)
   end
 
   def media_display_partial(file_set)
--- a/app/views/curation_concerns/base/_representative_media.html.erb
+++ b/app/views/curation_concerns/base/_representative_media.html.erb
@@ -1,5 +1,5 @@
-<% if presenter.representative_id.present? && (fs = ::FileSet.find(presenter.representative_id)) %>
-  <%= media_display fs %>
+<% if presenter.representative_id.present? %>
+  <%= media_display presenter.representative_presenter %>
 <% else %>
   <%= image_tag 'nope.png', class: "canonical-image" %>
 <% end %>
--- a/app/presenters/curation_concerns/work_show_presenter.rb
+++ b/app/presenters/curation_concerns/work_show_presenter.rb
@@ -40,6 +40,12 @@ module CurationConcerns
       @file_set_presenters ||= member_presenters(ordered_ids & file_set_ids)
     end
 
+    # @return FileSetPresenter presenter for the representative FileSets
+    def representative_presenter
+      return nil if representative_id.blank?
+      @representative_presenter ||= member_presenters([representative_id]).first
+    end
+
     # @return [Array<WorkShowPresenter>] presenters for the ordered_members that are not FileSets
     def work_presenters
       @work_presenters ||= member_presenters(ordered_ids - file_set_ids, work_presenter_class)
--- a/spec/factories/generic_works.rb
+++ b/spec/factories/generic_works.rb
@@ -23,6 +23,13 @@ FactoryGirl.define do
       end
     end
 
+    factory :work_with_representative_file do
+      before(:create) do |work, evaluator|
+        work.ordered_members << FactoryGirl.create(:file_set, user: evaluator.user, title: ['A Contained FileSet'])
+        work.representative_id = work.members[0].id
+      end
+    end
+
     factory :work_with_file_and_work do
       before(:create) do |work, evaluator|
         work.ordered_members << FactoryGirl.create(:file_set, user: evaluator.user)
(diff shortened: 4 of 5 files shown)


awead:
LGTM. Does a FileSet also use the same `_representative_media.html.erb` partial to display thumbnails?

cam156:
@awead no.  It does use the same helper method though, which is why I just changed to pass the presenter of the representative item in...  The FileSet does have all the information in its presenter to display the media display though.

awead:
@cam156 cool. Then I think this should be good to merge.

app/views/curation_concerns/base/_representative_media.html.erb
```diff
@@ -1,5 +1,5 @@
-<% if presenter.representative_id.present? && (fs = ::FileSet.find(presenter.representative_id)) %>
-  <%= media_display fs %>
+<% if presenter.representative_id.present? %>
```
jeremyf: This if statement and the `return nil if representative_id.blank?` of `#representative_presenter` are a duplication of knowledge. I'm wondering if removing the if statement and relying on the media_display method in the helper to ask the questions.

It would make testing the view logic easier (as we push it to the helper).

So in `_representative_media.html.erb`
```erb
<%= media_display(presenter.representative_presenter) %>
```

Then in `file_set_helper.rb`

```ruby
def media_display(presenter, locals = {})
  return image_tag('nope.png', class: "canonical-image") unless presenter.present?
  render(media_display_partial(presenter), locals.merge(file_set: presenter))
end
```

app/helpers/curation_concerns/file_set_helper.rb
```diff
@@ -7,9 +7,9 @@ def parent_path(parent)
     end
   end
 
-  def media_display(file_set, locals = {})
-    render media_display_partial(file_set),
-           locals.merge(file_set: file_set)
+  def media_display(presenter, locals = {})
```
jeremyf: Should this method name be `#file_set_media_display`? If so, for now, an alias could be added and deprecate the existing media_display.

The reason I ask is are there other `media_display` helper methods? If there are, these could be mixed into the same view and obliterate each other.

cam156: @jeremyf this media_display helper is being used on both the work and the file set display. for each we are passing a file_set presenter after this PR.  Before the PR it was getting a file set object or a file set presenter depending on where it was called from.  I changed the parameter to explicitly say we want a presenter.  I do not think it matters what type of presenter it is as long as it responds to the mime type methods (image? audio?...) and has a to_s method.

app/views/curation_concerns/base/_representative_media.html.erb
cam156: @jeremyf I hesitate to move the HTML into the helper.  it is harder for the UI people if the HTML is sprinkled in the helpers, presenters. I guess we could move the Image tag into it's own partial, or check to see if there is already another one and then just render that partial.

 I agree that the code is duplicated, but it is called from multiple places.

app/helpers/curation_concerns/file_set_helper.rb
jeremyf: @cam156 Ah. So perhaps the method is now in the wrong helper module? Like, should there be a media_display_helper module? That could be a follow-up refactor? Maybe add a TODO (or REVIEW) item in the comments?

app/views/curation_concerns/base/_representative_media.html.erb
jeremyf: @cam156 I can get behind your reasoning. So lets call this one "resolved".

app/helpers/curation_concerns/file_set_helper.rb
cam156: @jeremyf Good call.  I will add a review comment and an issue.

app/helpers/curation_concerns/file_set_helper.rb
```diff
@@ -7,9 +7,12 @@ def parent_path(parent)
     end
   end
 
-  def media_display(file_set, locals = {})
-    render media_display_partial(file_set),
-           locals.merge(file_set: file_set)
+  # REVIEW: Since this media display could theoretically work for
+  #         and object that inplements to_s and the Mime Type methos (image? audio? ...),
+  #         Should this really be in file_set or could it be in it's own helper class like media_helper?
```
cam156: @jeremyf Added a comment.  Will add and issue once the PR is merged.

app/helpers/curation_concerns/file_set_helper.rb
```diff
@@ -7,9 +7,12 @@ def parent_path(parent)
     end
   end
 
-  def media_display(file_set, locals = {})
-    render media_display_partial(file_set),
-           locals.merge(file_set: file_set)
+  # REVIEW: Since this media display could theoretically work for
```
mjgiarlo: Thoughts, @tpendragon @escowles @jcoyne?

I'm inclined to say :shipit: for now.

mjgiarlo:
Looks good to me. Tagged some folks to take a look at the `REVIEW` comment.

app/helpers/curation_concerns/file_set_helper.rb
escowles: I think this is fine as-is — pulling out to its own helper module would be a good refactor when there's something else that can use media_display.

PR #796, opened 2016-05-17
Generated code runs afoul of Rubocop.

mjgiarlo:
The Actor and Controller generated by the work generator violate the `Style/ClassAndModuleChildren` rule that is enabled by default in Rubocop.


jcoyne:
This is not a style issue as rubocop says, this code actually functions differently (where it searches for constants).  It doesn't matter in this example, so I don't care, but I think rubocop is wrong to say this is a style choice.

jeremyf:
The convention of Rubocop is that each layer of classes/modules are on separate lines and indented 2 spaces. Having used this convention, I much prefer it over the compact class/module definition that was previously there.

PR #798, opened 2016-05-17
Setup chain of jobs to ingest, characterize and create deriviates

### LINKED ISSUE #795: Job concurrency
### Descriptive summary

Since an original file must be present before characterization can be done, and the current `CharacterizeJob` also calls `CreateDerivativesJob`, there's a concurrency problem.

Should we make all three jobs run successively with a large superjob?
1. IngestLocalFileJob/IngestFileJob
2. CharacterizeJob
3. CreateDerivativesJob

Or do we want to explore some kind of locking?

Note, this was brought up with #793 because it will expose the problem.

awead:
Fixes #795 

FileActor calls IngestFileJob, which in turn calls CharacterizeJob and passes the local path to the file. CharacterizeJob in turn calls CreateDerivativesJob.

Also cleans-up and rearranges characterization code, adding an additional method to check for the availability of the characterization proxy file without needing to call Fedora.

ping @jeremyf The first commit addresses the `characterization_proxy?` method.


app/jobs/characterize_job.rb
```diff
@@ -2,9 +2,10 @@ class CharacterizeJob < ActiveJob::Base
   queue_as CurationConcerns.config.ingest_queue_name
 
   # @param [FileSet] file_set
-  # @param [String] filename a local path for the file to characterize. By using this, we don't have to pull a copy out of fedora.
+  # @param [String] filename a local path for the file to characterize so we don't have to pull a copy out of fedora.
   def perform(file_set, filename)
-    Hydra::Works::CharacterizationService.run(file_set, filename)
+    return unless file_set.characterization_proxy.content
```
mjgiarlo: Under what conditions will this be true? (Not a challenge, just curious.)

mjgiarlo:
@awead Looks like you need some stubbing to make sure fits doesn't try to run in Travis. Otherwise, I'd be happy to merge this PR. :+1:

app/jobs/characterize_job.rb
awead: @mjgiarlo I don't really know, honestly, but if there's no content then things would crash in weird ways. Perhaps we should raise an informative exception message?

mjgiarlo: If we're not sure how this situation might arise, I'm not sure we should be checking for it. If this short-circuit does get used, users are still in a weird spot, with files that are inexplicably without characterization or derivatives. I think we ought to understand the error condition better before handing it.

awead: @mjgiarlo yeah, I see that. Although, thinking about it more, I can see two cases in which we might want to catch this.

1. An implementation that uses a `characterization_proxy` other than `:original_file`

It'll be up to the user to make sure whatever relation their using gets ingested. They'd do this by passing it to `IngestFileJob`:

https://github.com/projecthydra-labs/curation_concerns/blob/master/app/jobs/ingest_file_job.rb#L9

But if they haven't set the fileset's characterization proxy to the same relation, then things would fail.

2. `IngestFileJob` fails for some reason

I'm not seeing anything in `IngestFileJob` that would stop it from calling the characterization job if something went awry. It's possible that `CreateDerivativesJob` would succeed, though. However, I can see putting an exception here as helpful to the user, and taking the "fail faster" approach instead of letting the job continue when we know at least one part of it will fail.

jeremyf: Do we want to be asking for the content? Won't that read the entire binary? Would it make sense to have a `file_set.characterization_proxy?` method or some such check that can prevent the more intimate knowledge of the workings of a characterization_proxy?

In reading the code, I see a guard of `file_set.characterization_proxy.content` followed by usage of `file_set.characterization_proxy`. This is a bit of a smell in that the guard has more intimate knowledge of the model than the service that is being called on the next line.

jeremyf: I followed the rabbit trail up to ActiveFedora::File and submitted an issue related to content? or some such thing. https://github.com/projecthydra/active_fedora/issues/1082

To resolve/answer my question, can you attach an `@todo Update characterization_proxy.content to not read content once https://github.com/projecthydra/active_fedora/issues/1082 is resolved`

awead: Yes, I see. Alternatively, we can check to see if the proxy is a `NullCharacterizationProxy`. This is what gets set if the relation (i.e. original_file) isn't there. Would it make sense to have the `characterization_proxy?` method in here:
https://github.com/projecthydra-labs/curation_concerns/blob/master/app/models/concerns/curation_concerns/characterization.rb#L17-30

jeremyf: I believe adding `characterization_proxy?` would be a great addition.

```ruby
def characterization_proxy?
  return true unless send(self.class.characterization_proxy).nil?
  false
end
```

PR #805, opened 2016-05-20
Return the first title for the page_title

mjgiarlo:
Fixes projecthydra/sufia#2052

Happy to have suggestions for testing strategies. Didn't see much coverage of this in the codebase already but perhaps I missed something obvious.

### DIFF
--- a/app/presenters/curation_concerns/work_show_presenter.rb
+++ b/app/presenters/curation_concerns/work_show_presenter.rb
@@ -23,7 +23,7 @@ module CurationConcerns
     end
 
     def page_title
-      solr_document.title
+      solr_document.title.first
     end
 
     # CurationConcern methods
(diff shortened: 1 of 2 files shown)


jcoyne:
Add a test in spec/presenters/curation_concerns/work_show_presenter_spec.rb, test for "foo" when solr_document.title returns ['foo', 'bar']

atz:
Definitely worth a test to codify expectation.

mjgiarlo:
:+1:

jeremyf:
Should we use `Array.wrap(solr_document.title).first`? I don't know how deep the assumption of what `#title` would look like.

awead:
Given its current position, `title` is pretty ubiquitous as an array, defined in `CurationConcerns::RequiredMetadata`. If it's truly "required" and is defined as multivalued, then I think we should always assume it's an array and not try to coax it into one because if it's somehow not an array, we've done something wrong and it should fail.

mjgiarlo:
Thx, all. Pushing my changes soon.

mjgiarlo:
@jeremyf Yeah, `#title` is already `Array.wrap`ped here: https://github.com/projecthydra/curation_concerns/blob/master/app/models/concerns/curation_concerns/solr_document_behavior.rb#L67

mjgiarlo:
@jcoyne @jeremyf @awead I believe this is ready.

PR #811, opened 2016-05-23
Change help text to always be visible

### LINKED ISSUE #777: Improve accessibility and usability of help text
### Descriptive summary

In our user and accessibility testing we found that the default presentation of help text is making some things difficult for users. Discussion for this issue went on at projecthydra/sufia#2003

justcolin:
Partially fixes #777, related to projecthydra/sufia#2003

Help text is now directly rendered to the page instead of being hidden behind a button/pop-up. This gets rid of a number of accessibility problems and simplifies the UI code.

The implementation ended up just involving the removal the custom code that was rendering the help buttons. After that, `simple_form`'s default rendering took over. The only problem was that the help text was being rendered after the input field, so a generator has been added to overwrite the default `simple_form` config files. All that this overwrite does is create identical initializers to what `simple_form` generates except the order of a couple of elements is flipped around. By just using the `simple_form` initializers users who want to customize things can use the `simple_form` DSL to make changes instead of overriding CC code.

The Visibility field will need to be cleaned up a bit now that all of that text is visible, but I have another pull request to take care of that. There is also probably some more code that can be removed, and I am investigating that now. I also have an upcoming pull request for Sufia which will help this CC change work better in Sufia.

* Delete custom help button generation code
* Add generator to create config files to make sure the help text is rendered how we want it and that it is easy to overwrite
* Remove old tests related to help button (no tests are needed for the new code since that would just be testing that `simple_form` works)


justcolin:
@projecthydra/sufia-ui-ux-advisors: here is how things look:

screen shot 2016-05-23 at 11 14 37 am

justcolin:
Restarting Travis since it was hanging after 45 minutes.

app/inputs/multi_value_with_help_input.rb
```diff
@@ -1,6 +1,4 @@
 class MultiValueWithHelpInput < MultiValueInput
```
jcoyne: If you're removing `WithHelpIcon` do we still need this class?

app/inputs/select_with_help_input.rb
```diff
@@ -1,3 +1,2 @@
 class SelectWithHelpInput < SimpleForm::Inputs::CollectionSelectInput
-  include WithHelpIcon
```
jcoyne: Still need this class?

app/inputs/select_with_modal_help_input.rb
```diff
@@ -1,11 +1,4 @@
 class SelectWithModalHelpInput < MultiValueWithHelpInput
```
jcoyne: Still need this class?

lib/generators/curation_concerns/install_generator.rb
```diff
@@ -104,5 +105,12 @@ def resource_types_config
     def simple_form_bootstrap
       generate 'simple_form:install --bootstrap'
     end
+
+    def simple_form_initializers
+      copy_file 'config/initializers/simple_form.rb',
```
jcoyne: How do these differ from the defaults distributed by simple_form? Why do we need them?

justcolin: They only differ by changing the order of the elements so the `hint` comes before the `input`. This is to make it match the design which was approved in projecthydra/sufia#2003 and because having the `hint` come after the `input` is worse for people using some assistive technologies.

app/inputs/select_with_modal_help_input.rb
justcolin: Probably not. I am currently investigating what views in CC and Sufia need to be altered when we remove these special classes.

app/inputs/select_with_help_input.rb
justcolin: Probably not. I am currently investigating what views in CC and Sufia need to be altered when we remove these classes.

app/inputs/multi_value_with_help_input.rb
justcolin: Probably not. I am currently investigating what views in CC and Sufia need to be altered when we remove these classes.

lib/generators/curation_concerns/install_generator.rb
jcoyne: That sounds good. Can you document the differences in the header of the templates?  Have you ticketed this recommendation in https://github.com/twbs/bootstrap?

lib/generators/curation_concerns/install_generator.rb
```diff
@@ -104,5 +105,12 @@ def resource_types_config
     def simple_form_bootstrap
       generate 'simple_form:install --bootstrap'
```
jcoyne: If you are providing the templates, there's no use in running this generator.

justcolin: It looks like that generator also generates some additional files. I assumed we wanted to let `simple_form` continue to generate these additional files:

- `config/locales/simple_form.en.yml`
- `lib/templates/erb/scaffold/_form.html.erb`

Would you rather we also generate those files even though it looks like we aren't overwriting them within CC?

lib/generators/curation_concerns/install_generator.rb
justcolin: I was going to wait until this ticket was sorted in CC and Sufia before trying to push the change upstream. I will show the differences in comments in the code.

lib/generators/curation_concerns/templates/config/initializers/simple_form.rb
```diff
@@ -0,0 +1,167 @@
...
+    # Calculates pattern from format validations for string inputs
+    b.optional :pattern
+
+    # Calculates min and max from length validations for numeric inputs
+    b.optional :min_max
+
+    # Calculates readonly automatically from readonly attributes
+    b.optional :readonly
+
+    ## Inputs
```
justcolin: CHANGED: the Inputs section used to start with `b.use :label_input`, now we put `b.use :input` at the bottom after the error.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,146 @@
...
+    b.use :html5
+    b.use :placeholder
+    b.optional :maxlength
+    b.optional :pattern
+    b.optional :min_max
+    b.optional :readonly
+    b.use :label, class: 'control-label'
+    b.use :error, wrap_with: { tag: 'span', class: 'help-block' }
+    b.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+    b.use :input, class: 'form-control'
```
justcolin: **CHANGED:** This line was moved down. it used to be at line 15.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,146 @@
...
+
+  config.wrappers :vertical_file_input, tag: 'div', class: 'form-group', error_class: 'has-error' do |b|
+    b.use :html5
+    b.use :placeholder
+    b.optional :maxlength
+    b.optional :readonly
+    b.use :label, class: 'control-label'
+    b.use :error, wrap_with: { tag: 'span', class: 'help-block' }
+    b.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+    b.use :input
```
justcolin: CHANGED: This line was moved down. it used to be at line 26.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,146 @@
...
+    b.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+  end
+
+  config.wrappers :vertical_radio_and_checkboxes, tag: 'div', class: 'form-group', error_class: 'has-error' do |b|
+    b.use :html5
+    b.optional :readonly
+    b.use :label, class: 'control-label'
+    b.use :error, wrap_with: { tag: 'span', class: 'help-block' }
+    b.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+    b.use :input
```
justcolin: CHANGED: This line was moved down. it used to be at line 47.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,146 @@
...
+    b.optional :maxlength
+    b.optional :pattern
+    b.optional :min_max
+    b.optional :readonly
+    b.use :label, class: 'col-sm-3 control-label'
+
+    b.wrapper tag: 'div', class: 'col-sm-9' do |ba|
+      ba.use :error, wrap_with: { tag: 'span', class: 'help-block' }
+      ba.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+      ba.use :input, class: 'form-control'
```
justcolin: **CHANGED:** This line was moved down. it used to be at line 62.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,146 @@
...
+    b.use :html5
+    b.use :placeholder
+    b.optional :maxlength
+    b.optional :readonly
+    b.use :label, class: 'col-sm-3 control-label'
+
+    b.wrapper tag: 'div', class: 'col-sm-9' do |ba|
+      ba.use :error, wrap_with: { tag: 'span', class: 'help-block' }
+      ba.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+      ba.use :input
```
justcolin: **CHANGED:** This line was moved down. it used to be at line 76.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,146 @@
...
+  config.wrappers :horizontal_radio_and_checkboxes, tag: 'div', class: 'form-group', error_class: 'has-error' do |b|
+    b.use :html5
+    b.optional :readonly
+
+    b.use :label, class: 'col-sm-3 control-label'
+
+    b.wrapper tag: 'div', class: 'col-sm-9' do |ba|
+      ba.use :error, wrap_with: { tag: 'span', class: 'help-block' }
+      ba.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+      ba.use :input
```
justcolin: **CHANGED:** This line was moved down. it used to be at line 103.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+  config.wrappers :multi_select, tag: 'div', class: 'form-group', error_class: 'has-error' do |b|
+    b.use :html5
+    b.optional :readonly
+    b.use :label, class: 'control-label'
+    b.wrapper tag: 'div', class: 'form-inline' do |ba|
+      ba.use :error, wrap_with: { tag: 'span', class: 'help-block' }
+      ba.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+      ba.use :input, class: 'form-control'
```
justcolin: **CHANGED:** This line was moved down. it used to be at line 127.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,146 @@
...
+    b.use :html5
+    b.use :placeholder
+    b.optional :maxlength
+    b.optional :pattern
+    b.optional :min_max
+    b.optional :readonly
+    b.use :label, class: 'sr-only'
+    b.use :error, wrap_with: { tag: 'span', class: 'help-block' }
+    b.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+    b.use :input, class: 'form-control'
```
justcolin: **CHANGED:** This line was moved down. it used to be at line 117.

justcolin:
I am currently working on altering the views so I can pull out the classes that @jcoyne mentioned.

lib/generators/curation_concerns/install_generator.rb
jcoyne: I don't think we'll need those. Let's do away with this.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
jcoyne: I mean, can you document this in the code.  Just put a comment at the top of the file like "this differs from the default template in that we moved the hints above the controls for accessibility"

lib/generators/curation_concerns/install_generator.rb
justcolin: `simple_form.en.yml` is useful to give a dedicated place for adding simple_form settings and we use it in a few projects. If we remove this line we may soon have to at least add another generator to generate that file.

justcolin: I'll check to see if simple_form complains without those files.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
justcolin: Sorry for he misunderstanding, will do.

jonearley:
screen shot 2016-05-24 at 10 26 51 am
screen shot 2016-05-24 at 10 28 51 am
screen shot 2016-05-24 at 10 29 02 am

These colors are bootstrap defaults: danger, info, and primary.

@projecthydra/sufia-ui-ux-advisors thought and suggestions on what the default required tag color should be?

lib/generators/curation_concerns/install_generator.rb
justcolin: Things appear to work fine without those files so I removed that line and changed the `desc`.

lib/generators/curation_concerns/templates/config/initializers/simple_form_bootstrap.rb
justcolin: Resolved.

lib/generators/curation_concerns/install_generator.rb
```diff
@@ -101,8 +101,11 @@ def resource_types_config
       copy_file "config/authorities/resource_types.yml", "config/authorities/resource_types.yml"
     end
 
-    def simple_form_bootstrap
-      generate 'simple_form:install --bootstrap'
```
awead: You could keep this generator in here, then copy over the generated files. This ensures if later versions of simple form install additional files, they will be generated.

justcolin: I am okay with either option. @jcoyne asked me to remove this line, so what is the majority opinion? @mjgiarlo, @cam156, etc?

mjgiarlo: I can see it either way, though I have a slight preference for leaving it as is.

When simple_form changes underneath us, we should change. I'm loath to guess at how it will change. For instance, what if they decide to break simple_form.rb into two files with different names? Then we copy over two files that potentially aren't used anymore and may be overridden in part or in full by the new files.

awead: 👍  and we've pinned it to ~> 3.2 so major versions won't be included.

jcoyne: I just added this a few weeks ago. It's not something sufia 6 did. So clearly we don't need it.

PR #812, opened 2016-05-23
Fix inactive rights editing bug

### LINKED ISSUE #775: Edit Work that has inactive rights.
### Descriptive summary
Editing a work that has inactive rights will force a user to either cancel the edits or select a rights statement from the list of active rights.

CC HEAD.

### Expected behavior
Edit the metadata of a work without having to choose a new rights statement.

### Actual behavior
Editting the metadata requires selection of a currently active rights statement.  In the case of legacy Works, this will require the user to change the rights from what was originally granted for the Work.

1. Create Work
1. Set rights selection to an rights statement with active: false. (via rails console)
1. Edit the Work

### Related work
This is a consequence of #767.  The contributor neglected to bring this issue to light or have a solution for it at the time of the PR.  Not going to name any names, but it was me.

jenlindner:
Fixes #775 

Creates an html select option containing data-attributes with the work's inactive rights, which the javascript then uses to add as options into the work edit form's rights remove selects. The user can save their edits without altering their rights or remove the inactive ones.

* the addition of data-attributes and classes to the html that is used to create rights selectors
* the addition of javascript that looks for those attributes and classes and creates options for the work's inactive rights, with a class on the options to be used at the app level to indicate the rights are inactive (if desired)

### DIFF
--- a/app/inputs/select_with_modal_help_input.rb
+++ b/app/inputs/select_with_modal_help_input.rb
@@ -31,6 +31,16 @@ class SelectWithModalHelpInput < MultiValueWithHelpInput
       @rendered_first_element = true
 
       html_options.merge!(options.slice(:include_blank))
+      html_options = add_force_option_data_to_html_options(value, html_options)
       template.select_tag(attribute_name, template.options_for_select(select_options, value), html_options)
+    end
+
+    def add_force_option_data_to_html_options(value, html_options)
+      return html_options if value.blank? || RightsService.active?(value)
+      html_options.tap do |opts|
+        opts[:class] << ' force-select'
+        opts[:'data-force-label'] = RightsService.label(value)
+        opts[:'data-force-value'] = value
+      end
     end
 end
--- a/app/assets/javascripts/curation_concerns/curation_concerns.js
+++ b/app/assets/javascripts/curation_concerns/curation_concerns.js
@@ -8,7 +8,7 @@
 //= require curation_concerns/file_manager/member
 //= require curation_concerns/batch_select
 //= require curation_concerns/collections
-
+//= require curation_concerns/rights
 
 // Initialize plugins and Bootstrap dropdowns on jQuery's ready event as well as
 // Turbolinks's page change event.
--- a/app/assets/javascripts/curation_concerns/rights.js
+++ b/app/assets/javascripts/curation_concerns/rights.js
@@ -1,0 +1,8 @@
+Blacklight.onLoad(function () {
+  $('.force-select').each(function(){
+    if ($(this).data('force-value')){
+      var option = $('<option>', { value: $(this).data('force-value'), text: $(this).data('force-label')});
+      $(this).append($(option).attr('selected','selected'));
+    }
+  });
+});
--- a/app/services/rights_service.rb
+++ b/app/services/rights_service.rb
@@ -2,8 +2,22 @@ module RightsService
   mattr_accessor :authority
   self.authority = Qa::Authorities::Local.subauthority_for('rights')
 
-  def self.select_options
+  def self.select_all_options
+    authority.all.map do |element|
+      [element[:label], element[:id]]
+    end
+  end
+
+  def self.select_active_options
     active_elements.map { |e| [e[:label], e[:id]] }
+  end
+
+  def self.active?(id)
+    authority.find(id).fetch('active')
+  end
+
+  def self.select_inactive_options
+    inactive_elements.map { |e| [e[:label], e[:id]] }
   end
 
   def self.label(id)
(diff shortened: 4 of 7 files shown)


app/inputs/select_with_modal_help_input.rb
```diff
@@ -31,6 +31,13 @@ def build_field(value, _index)
       @rendered_first_element = true
 
       html_options.merge!(options.slice(:include_blank))
-      template.select_tag(attribute_name, template.options_for_select(select_options, value), html_options)
+      render_options = select_options
```
jeremyf: Is there a reason for copying the variable?

jenlindner: Just the added clarity that there will be additional options rendered.

app/inputs/select_with_modal_help_input.rb
```diff
@@ -31,6 +31,13 @@ def build_field(value, _index)
...
+      if value.present? && render_options.find { |opt| opt[1] == value }.nil?
+        force_option = RightsService.select_inactive_options.find { |opt| opt[1] == value } || [value, value]
```
jeremyf: I'm wondering about factoring lines 35 and 36 into a service method? Two sets of find in view logic is a bit of a warning flag for me. There is also an assumption about the found results (i.e. `opt[1]` implies we get back an array).

One option I had thought of was:

```ruby
RightsService.with_forced_label_and_value_for(render_options, value) do | label, value |
  html_options[:class] << "force-select"
  html_options[:'data-force-label'] = label
  html_options[:'data-force-value'] = value
end
```

I don't know if that is more clarifying or less from a view legibility stand point?

jenlindner: You have a point, I have a thought. I'll post it when I get in to work.

app/inputs/select_with_modal_help_input.rb
mbklein: Also because `select_options` is a method, not a variable.

mbklein: D'oh! Never mind. The variable is left over from a previous iteration of the fix where the forced option was added to the select at render time, not at runtime.

app/inputs/select_with_modal_help_input.rb
jenlindner: It doesn't belong in the Rights Service, because that's concerned only with what rights are currently active or inactive in general, not with which are associated with works. And it's not exactly view logic (occurring in the view itself), because this class is a view helper already. But, the extra logic there could definitely be more cleanly handled in another private method on that class, like so:
```
    def add_force_option_data_to_html_options(select_options, value, html_options)
      if value.present? && select_options.find { |opt| opt[1] == value }.nil?
        force_option = RightsService.select_inactive_options.find { |opt| opt[1] == value } || [value, value]
        html_options[:class] << "force-select"
        html_options[:'data-force-label'] = force_option[0]
        html_options[:'data-force-value'] = force_option[1]
      end
      html_options
    end
```

and then in the build_field method you get this:

```
      html_options.merge!(options.slice(:include_blank))
      html_options = add_force_option_data_to_html_options(select_options, value, html_options)
      template.select_tag(attribute_name, template.options_for_select(select_options, value), html_options)
```

I like this, what do you think?

jeremyf: I can get behind that.

jenlindner:
I like a refactor MBK did with part of this, I included it and am running the tests now. Will push it up in a minute.

mbklein:
👍

app/inputs/select_with_modal_help_input.rb
```diff
@@ -31,6 +31,18 @@ def build_field(value, _index)
       @rendered_first_element = true
 
       html_options.merge!(options.slice(:include_blank))
+      html_options = add_force_option_data_to_html_options(value, html_options)
       template.select_tag(attribute_name, template.options_for_select(select_options, value), html_options)
     end
+
+    def add_force_option_data_to_html_options(value, html_options)
```
mjgiarlo: Can you rewrite this using a guard clause?

```ruby
def add_force_option_data_to_html_options(value, html_options)
  return html_options unless value.present? || RightsService.active?(value) # er, check my boolean logic here...
  html_options.tap do |opts|
    opts[:class] << ' force-select'
    opts[:'data-force-label'] = RightsService.label(value)
    opts[:'data-force-value'] = value
  end
end
```

app/inputs/select_with_modal_help_input.rb
```diff
@@ -31,6 +31,18 @@ def build_field(value, _index)
...
+      if value.present?
+        unless RightsService.active?(value)
+          html_options[:class] << "force-select"
```
mjgiarlo: Does `'force-select'` need a space in the string so that you don't get, e.g., `<span class="foo bar bazforce-select">` ?

mjgiarlo:
@jenlindner Can you rebase this PR?

mbklein:
Superseded by #817 after a merge conflict with current master

PR #817, opened 2016-05-25
Fix inactive rights editing bug, Take two

### LINKED ISSUE #775: Edit Work that has inactive rights.

mbklein:
Fixes #775 ; refs #811 and #812

Allows a user to save their edits without altering their rights or removing the inactive ones.

* the addition of a `MultiValueSelectInput` input class that presents individual drop-down select elements for multi-valued fields 
* the addition of a helper that ensures that current values are always selectable, even if marked inactive by the controlling authority

### DIFF
--- a/app/services/rights_service.rb
+++ b/app/services/rights_service.rb
@@ -2,8 +2,18 @@ module RightsService
   mattr_accessor :authority
   self.authority = Qa::Authorities::Local.subauthority_for('rights')
 
-  def self.select_options
+  def self.select_all_options
+    authority.all.map do |element|
+      [element[:label], element[:id]]
+    end
+  end
+
+  def self.select_active_options
     active_elements.map { |e| [e[:label], e[:id]] }
+  end
+
+  def self.active?(id)
+    authority.find(id).fetch('active')
   end
 
   def self.label(id)
--- a/app/helpers/curation_concerns/main_app_helpers.rb
+++ b/app/helpers/curation_concerns/main_app_helpers.rb
@@ -9,4 +9,5 @@ module CurationConcerns::MainAppHelpers
   include CurationConcerns::LeaseHelper
   include CurationConcerns::CollectionsHelper
   include CurationConcerns::PermissionsHelper
+  include CurationConcerns::RightsHelper
 end
--- a/app/helpers/curation_concerns/rights_helper.rb
+++ b/app/helpers/curation_concerns/rights_helper.rb
@@ -1,0 +1,9 @@
+module CurationConcerns::RightsHelper
+  def include_current_value(value, _index, render_options, html_options)
+    unless value.blank? || RightsService.active?(value)
+      html_options[:class] << ' force-select'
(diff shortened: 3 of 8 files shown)


mbklein:
@projecthydra/sufia-ui-ux-advisors This PR keeps the look and feel of the form as they were before #811, while fixing the bug it was designed to fix. @grosscol suggested that you might want to advise on whether multi-valued controlled fields would better be served by a single, HTML-standard `<select multiple="true">` component.

app/services/rights_service.rb
```diff
@@ -2,15 +2,33 @@ module RightsService
...
   def self.label(id)
     authority.find(id).fetch('term')
   end
 
   def self.active_elements
     authority.all.select { |e| authority.find(e[:id])[:active] }
   end
+
+  def self.inactive_elements
+    authority.all.select { |e| authority.find(e[:id])[:active] == false }
```
jcoyne: Is there some reason you aren't doing `.reject { |e| authority.find(e[:id])[:active] }`?

app/services/rights_service.rb
```diff
@@ -2,15 +2,33 @@ module RightsService
...
+
+  def self.select_active_options
     active_elements.map { |e| [e[:label], e[:id]] }
   end
 
+  def self.active?(id)
+    authority.find(id).fetch('active')
+  end
+
+  def self.select_inactive_options
```
jcoyne: Is this method ever used? If not, can we get rid of it?

app/services/rights_service.rb
justcolin: Sorry, accidentally deleted a comment by @jcoyne. He said:

Is there some reason you aren't doing `.reject { |e| authority.find(e[:id])[:active] }`?

app/services/rights_service.rb
mbklein: Ah, it's a remnant of the original fix that got refactored out. It can go.

app/services/rights_service.rb
mbklein: Fair point. :+1:

mbklein:
We'll also fix the rubocop complaints before we push again.

justcolin:
The `MultiValueSelectInput` stuff looks good to me.

mjgiarlo:
Looks good to me! @tpendragon @jcoyne @escowles

PR #818, opened 2016-05-26
Improved UI for required field labels

jonearley:
Before:
screen shot 2016-05-26 at 10 01 01 am

After:
screen shot 2016-05-24 at 10 28 51 am

Related discussion: https://github.com/projecthydra/sufia/issues/2076

### DIFF
--- a/config/locales/curation_concerns.en.yml
+++ b/config/locales/curation_concerns.en.yml
@@ -159,6 +159,8 @@ en:
   file_manager:
     link_text: 'File Manager'
   simple_form:
+    required:
+      html: '<span class="label label-info required-tag">required</span>'
     hints:
       defaults:
         description: 'Please keep your description to 300 words or fewer.'
--- a/lib/generators/curation_concerns/templates/config/initializers/simple_form.rb
+++ b/lib/generators/curation_concerns/templates/config/initializers/simple_form.rb
@@ -102,7 +102,7 @@ SimpleForm.setup do |config|
   # config.item_wrapper_class = nil
 
   # How the label text should be generated altogether with the required text.
-  # config.label_text = lambda { |label, required, explicit_label| "#{required} #{label}" }
+  config.label_text = ->(label, required, _) { "#{label} #{required}" }
 
   # You can define the class to use on all labels. Default is nil.
   # config.label_class = nil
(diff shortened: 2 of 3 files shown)


config/locales/curation_concerns.en.yml
```diff
@@ -159,6 +159,8 @@ en:
   file_manager:
     link_text: 'File Manager'
   simple_form:
+    required:
+      html: '<span class="label label-info required-tag">required</span>'
```
jcoyne: Can we move the html into the simple_form initializer? That wouldn't change even if we're using a different language.

jonearley: After looking into this a bit more simple_form has the attribute element built in. Since we're no longer using the "*" it's use wouldn't make semantic sense. https://github.com/plataformatec/simple_form/blob/ac63da0f0b38df82385493f25f649e320b0acb4a/lib/simple_form/components/labels.rb#L10

The documented/simple_form way to use custom required markup is to define html in the yaml file. https://github.com/plataformatec/simple_form/blob/ac63da0f0b38df82385493f25f649e320b0acb4a/lib/generators/simple_form/templates/config/locales/simple_form.en.yml#L8-L10

Considering the trade-offs I think this implementation is the "right way" even though the defined yaml html will stay the same with different languages. We can also avoid writing custom code this way.

jcoyne: 👌

lib/generators/curation_concerns/templates/config/initializers/simple_form.rb
```diff
@@ -102,7 +102,7 @@
   # config.item_wrapper_class = nil
 
   # How the label text should be generated altogether with the required text.
-  # config.label_text = lambda { |label, required, explicit_label| "#{required} #{label}" }
+  config.label_text = lambda { |label, required| "#{label} #{required}" }
```
mjgiarlo: Change to:

```ruby
config.label_text = ->(label, required) { "#{label} #{required}" }
```

jonearley: @mjgiarlo thanks! I pushed this change.

PR #819, opened 2016-05-26
Add FacetedAttributeRenderer

mbklein:
Prerequisite for fixing [#2081@sufia](https://github.com/projecthydra/sufia/issues/2081)

Add a `FacetedAttributeRenderer` to generate facet-type linked attributes

The `LinkedAttributeRenderer` generates links in the `/catalog?q=value&search_field=field_name` format. These links don't work with Sufia's default configuration. The faceted link type (`/catalog?f[field_name][]=value`) does work.

* Add `FacetedAttributeRenderer` to generate the new link types
* Update `PresentsAttributes` to accept a `:render_as` option for more flexible specification of which attribute renderer to use

### DIFF
--- a/app/presenters/curation_concerns/presents_attributes.rb
+++ b/app/presenters/curation_concerns/presents_attributes.rb
@@ -4,7 +4,8 @@ module CurationConcerns
     # Present the attribute as an HTML table row.
     #
     # @param [Hash] options
-    # @option options [TrueClass, FalseClass] :catalog_search_link return a link to a catalog search for that text if true
+    # @option options [Symbol] :render_as use an alternate renderer
+    #   (e.g., :linked or :linked_attribute to use LinkedAttributeRenderer)
     # @option options [String] :search_field If the method_name of the attribute is different than
     #   how the attribute name should appear on the search URL,
     #   you can explicitly set the URL's search field name
@@ -39,11 +40,24 @@ module CurationConcerns
 
     private
 
-      def renderer_for(field, options)
-        if options[:catalog_search_link]
-          Renderers::LinkedAttributeRenderer
-        elsif field == :rights
-          Renderers::RightsAttributeRenderer
+      def find_renderer_class(name)
+        renderer = nil
+        ['Renderer', 'AttributeRenderer'].each do |suffix|
+          const_name = "#{name.to_s.camelize}#{suffix}".to_sym
+          renderer = begin
+            Renderers.const_get(const_name)
+          rescue NameError
+            nil
+          end
+          break unless renderer.nil?
+        end
+        raise NameError, "unknown renderer type `#{name}`" if renderer.nil?
+        renderer
+      end
+
+      def renderer_for(_field, options)
+        if options[:render_as]
+          find_renderer_class(options[:render_as])
         else
           Renderers::AttributeRenderer
         end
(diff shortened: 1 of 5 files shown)


app/presenters/curation_concerns/presents_attributes.rb
```diff
@@ -39,10 +39,24 @@ def microdata_type_to_html
...
       def renderer_for(field, options)
-        if options[:catalog_search_link]
+        if options[:render_as]
+          # Either `render: :linked` and `render: :linked_attribute` will use the LinkedAttributeRenderer
+          find_renderer_class(options[:render_as])
+        elsif options[:catalog_search_link]
+          # Should we deprecate `catalog_search_link: true` in favor of `render_as: :linked`?
           Renderers::LinkedAttributeRenderer
         elsif field == :rights
+          # Should we deprecate this in favor of `render_as: :rights`?
```
jcoyne: yes. Also no need to deprecate. Just make the change.

app/presenters/curation_concerns/presents_attributes.rb
```diff
@@ -39,10 +39,24 @@ def microdata_type_to_html
...
+        renderer
+      end
+
       def renderer_for(field, options)
-        if options[:catalog_search_link]
+        if options[:render_as]
+          # Either `render: :linked` and `render: :linked_attribute` will use the LinkedAttributeRenderer
+          find_renderer_class(options[:render_as])
+        elsif options[:catalog_search_link]
+          # Should we deprecate `catalog_search_link: true` in favor of `render_as: :linked`?
```
jcoyne: yes

mbklein:
Yep. That was dumb. Force-push fix on the way.

app/presenters/curation_concerns/presents_attributes.rb
```diff
@@ -39,11 +40,24 @@ def microdata_type_to_html
...
+            nil
+          end
+          break unless renderer.nil?
+        end
+        raise NameError, "unknown renderer type `#{name}`" if renderer.nil?
+        renderer
+      end
+
+      def renderer_for(_field, options)
+        if options[:render_as]
```
mjgiarlo: What would you think about rewriting this as a guard clause? Like:

```ruby
return Renderers::AttributeRenderer unless options.key? :render_as
find_renderer_class(options[:render_as])
```

app/presenters/curation_concerns/presents_attributes.rb
```diff
@@ -39,11 +40,24 @@ def microdata_type_to_html
 
     private
 
-      def renderer_for(field, options)
-        if options[:catalog_search_link]
-          Renderers::LinkedAttributeRenderer
-        elsif field == :rights
-          Renderers::RightsAttributeRenderer
+      def find_renderer_class(name)
```
mjgiarlo: Can we use a short-circuit pattern here instead of the loop break and keeping track of whether `renderer` is `nil` or not?

```ruby
def find_renderer_class(name)
    [ 'Renderer', 'AttributeRenderer'].each do |base|
    const_name = "#{name.to_s.camelize}#{base}"
    return Renderers.const_get(const_name) if Renderers.const_defined?(const_name)
  end
  raise NameError, "unknown renderer type `#{name}`"
end
```

mbklein: I fell into a trap earlier in that `#const_get` seems to trigger autoloading, while `#const_defined?` doesn't. I couldn't figure out a reasonable way around it that didn't depend on `begin ... rescue NameError`

PR #820, opened 2016-05-26
Keep solr_document logic out of the presenter

jcoyne:

Only the solr_document class should be aware of the name of the solr
fields.


mjgiarlo:
:red_circle: :crying_cat_face:

PR #822, opened 2016-05-27
Elminiate the AssignIdentifierActor. It's unnecessary

jcoyne:


barmintor:
The noid service is called directly via override from ActiveFedora.

PR #829, opened 2016-05-31
Contextual Nested Works

tpendragon:
Adds nested works, but I have some questions about how far to go:

Features in PR
============

1. "Attach Child Generic Work" button created.
2. Configurable settings for which items can be children.
3. File Manager can reorder/set labels for both FileSets and Works.
4. "Files" display is now "Members" display.

Potential Features (AKA should I try to add these?)
==============

1. Breadcrumb Paths (Does this mean a recursive show path for a work with works with works?)
2. Thumbnail Selection (Allow a child Generic Work to be representative - so its thumbnail is the parent's thumbnail)
3. Default Discovery (Should children be discoverable? Or just leave this up to implementors?)
4. Indexing of parents (Necessary for filtering discovery, but may be an upstream concern?)

### DIFF
--- a/app/actors/curation_concerns/actors/add_to_work_actor.rb
+++ b/app/actors/curation_concerns/actors/add_to_work_actor.rb
@@ -1,0 +1,35 @@
+module CurationConcerns
+  module Actors
+    class AddToWorkActor < AbstractActor
+      def create(attributes)
+        work_ids = attributes.delete(:in_works_ids)
+        next_actor.create(attributes) && add_to_works(work_ids)
+      end
+
+      def update(attributes)
+        work_ids = attributes.delete(:in_works_ids)
+        add_to_works(work_ids) && next_actor.update(attributes)
+      end
+
+      private
+
+        def add_to_works(new_work_ids)
+          return true unless new_work_ids.present?
+          (curation_concern.in_works_ids - new_work_ids).each do |old_id|
+            work = ::ActiveFedora::Base.find(old_id)
+            work.ordered_members.delete(curation_concern)
+            work.members.delete(curation_concern)
+            work.save
+          end
+
+          # add to new
+          (new_work_ids - curation_concern.in_works_ids).each do |work_id|
+            work = ::ActiveFedora::Base.find(work_id)
+            work.ordered_members << curation_concern
+            work.save
+          end
+          true
+        end
+    end
+  end
+end
--- a/app/models/concerns/curation_concerns/nested_works.rb
+++ b/app/models/concerns/curation_concerns/nested_works.rb
@@ -1,0 +1,14 @@
+module CurationConcerns
+  module NestedWorks
+    extend ActiveSupport::Concern
+
+    included do
+      class_attribute :valid_child_concerns
+      self.valid_child_concerns = CurationConcerns::ClassifyConcern.new.all_curation_concern_classes
+    end
+
+    def in_works_ids
+      in_works.map(&:id)
+    end
+  end
+end
--- a/app/views/curation_concerns/base/_member.html.erb
+++ b/app/views/curation_concerns/base/_member.html.erb
@@ -2,11 +2,11 @@
   <td class="thumbnail">
     <%= render_thumbnail_tag member %>
   </td>
-  <td class="attribute filename"><%= link_to(member.link_name, main_app.curation_concerns_file_set_path(member)) %></td>
-  <td class="attribute date_uploaded"><%= member.date_uploaded %></td>
+  <td class="attribute filename"><%= link_to(member.link_name, contextual_path(member, @presenter)) %></td>
+  <td class="attribute date_uploaded"><%= member.try(:date_uploaded) %></td>
   <td class="attribute permission"><%= member.permission_badge %></td>
   <td>
-    <%= render 'curation_concerns/file_sets/actions', file_set: member %>
+    <%= render 'actions', member: member %>
   </td>
 </tr>
 
--- a/app/views/curation_concerns/base/_related_files.html.erb
+++ b/app/views/curation_concerns/base/_related_files.html.erb
@@ -1,7 +1,7 @@
-<% if presenter.file_set_presenters.present? %>
+<% if presenter.member_presenters.present? %>
 <div class="panel panel-default related_files">
   <div class="panel-heading">
-    <h2>Files</h2>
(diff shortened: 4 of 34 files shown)


escowles:
IMHO:

1. Breadcrumbs would be great, but I don't think we need to recurse.  Just having the parent's breadcrumb seems like enough for now.
2. Thumbnail selection seems good, but fine to make it a separate ticket.
3. I think the children should be discoverable.  I wouldn't want to have something like the suppressing/joining logic we have in Plum, since different implementers might have pretty different ideas about what should be discovered.
4. Ditto: this is pretty tightly tied to number 3.

vantuyls:
Sweet. So, to your questions - Im looking at this from a Sufia perspective, so keep that in mind. 

1. Breadcrumb Paths (Does this mean a recursive show path for a work with works with works?)
**I would like to see this implemented**
2. Thumbnail Selection (Allow a child Generic Work to be representative - so its thumbnail is the parent's thumbnail)
**This is less important from my perspective, but doesn't CC do this now for collections?**
3. Default Discovery (Should children be discoverable? Or just leave this up to implementors?)
**Huge question - YUGE. I'm not exactly sure where I fall on this, but my knee-jerk is to say that children should be discoverable. I really wonder, though, what the UI looks like for discovery of these multi-part works as both the parent and child works are potentially important in a result set. This also raises issues about consistency of representation of types of content in the repository in order for discoverability to feel consistent, but thats sort of out of scope for this issue. **
4. Indexing of parents (Necessary for filtering discovery, but may be an upstream concern?)
**see above**

tpendragon:
So regarding breadcrumb paths, I know Sufia does this already - how much would I be screwing up by doing it here? @jcoyne?

app/models/concerns/curation_concerns/nested_works.rb
```diff
@@ -0,0 +1,14 @@
+module CurationConcerns
+  module NestedWorks
+    extend ActiveSupport::Concern
+
+    included do
+      class_attribute :valid_child_concerns
+      self.valid_child_concerns = CurationConcerns::ClassifyConcern.new.all_curation_concern_classes
```
escowles: I would definitely like to see example config for this included somewhere (in the main curation_concerns initializer?).  Something like:

```
# By default, any work type is a valid child.  To override to limit to certain types, set valid_child_concerns
# Book.valid_child_concerns = [Page]
```

jpstroop:
1. Breadcrumb Paths - Agree w/ Esmé, just link to the parent. Might be nice to make it configurable, but start there.
 2. Thumbnail Selection - Yes, draw from a child by default. Does recursion make this difficult (do you need to support getting a thumbnail from a child's child?)
 3. Default Discovery - Meh. If others want children to be discoverable, sure. Configurable? Per work?

app/views/curation_concerns/base/_member.html.erb
```diff
@@ -1,12 +1,12 @@
-<tr class="<%= dom_class(member) %> attributes">
+<filestr class="<%= dom_class(member) %> attributes">
```
mjgiarlo: Typo?

app/views/curation_concerns/base/_related_files.html.erb
```diff
@@ -14,11 +14,11 @@
...
     <tbody>
-      <%= render partial: 'member', collection: presenter.file_set_presenters %>
+      <%= render partial: 'member', collection: presenter.member_presenters %>
     </tbody>
   </table>
 </div>
 <% elsif can? :edit, presenter.id %>
-  <h2>Files</h2>
-  <p class="text-center"><em>This <%= presenter.human_readable_type %> has no files associated with it. You can add one using the "Attach a File" button below.</em></p>
+  <h2>Members</h2>
```
mjgiarlo: `Members` makes sense to me in this context, but it could (?) trip up end-users. Thoughts, @projecthydra/sufia-ui-ux-advisors ?

We recently merged a PR in Sufia where we referred to an indeterminate thing that could be either a Work or a Collection as an "item," FWIW.

cbeer: Commit to i18n and make it a downstream problem?

vantuyls:
> We recently merged a PR in Sufia where we referred to an indeterminate thing that could be either a Work or a Collection as an "item," FWIW.

is that thing indeterminate as a Work or Collection because child things haven't been assigned to it yet (ie it's an empty thing)? I threw up in my mouth when i saw that you called something an item.

mjgiarlo:
@vantuyls :speech_balloon: 
> is that thing indeterminate as a Work or Collection because child things haven't been assigned to it yet (ie it's an empty thing)? I threw up in my mouth when i saw that you called something an "item".

No, and as I reply, I realize we opted *against* that approach. The issue was that we were rolling up locale translations for collections and works into a single block, and rather than do that, we broke them out into largely duplicative translations precisely because we thought it was important to distinguish works from collections and to avoid new terminology ("items"). Moving on...

app/models/concerns/curation_concerns/nested_works.rb
tpendragon: Maybe just add this to the top of every generated CC?

tpendragon:
@mjgiarlo @vantuyls https://vimeo.com/169123709 Quick demo of what this PR adds.

app/actors/curation_concerns/actors/add_to_work_actor.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      def update(attributes)
+        work_ids = attributes.delete(:in_works_ids)
+        add_to_works(work_ids) && next_actor.update(attributes)
+      end
+
+      private
+
+        def add_to_works(new_work_ids)
+          return true unless new_work_ids.present?
```
revgum: I discovered that when we try to remove all work_ids from an existing work, this line prevents the update from functioning. We'd like to be able to "remove all parent works from a child work" by posting an empty array of :in_works_ids .. Should this line be removed entirely, or does it serve a particular purpose?

jcoyne: I'm guessing this was intended to only be a nil check

tpendragon: Yes, was meant to be a nil check.

PR #832, opened 2016-06-01
Fix issue with rights and source attributes

### LINKED ISSUE #831: Rights and source are not permitted
The latest version of curation_concerns rejects the source and rights parameters when POSTing a new work or updating. The log shows:

`Unpermitted parameters: source, rights`

This occurs in the v1.0.0.beta5 release as well as on the master branch. I imagine this has something to do with the changes in #817?

##### Steps to reproduce the behavior
1. Create a new work
2. Assign rights and source on the form
3. Feel sad when they aren't saved

Adding
```
fill_in('Source', with: source)
select 'Attribution 3.0 United States', from: 'generic_work[rights][]'
```
and
```
expect(page).to have_content(source)
expect(page).to have_link 'Attribution-ShareAlike 3.0 United States', href: 'http://creativecommons.org/licenses/by-sa/3.0/us/'
```
to `spec/features/create_work_spec.rb` causes the test to fail.

eliotjordan:
Fixes #831 

Corrects issue with unpermitted source and rights parameters.

The rights field became multi-valued in #817, but the overridden `multiple?` method in `forms/work_form` forced the param to be single-valued. The source field was not included in the terms array, solr document, or in the attribute_rows view partial. It was added and now both attributes can be saved and viewed on the show page.

### DIFF
--- a/app/forms/curation_concerns/forms/work_form.rb
+++ b/app/forms/curation_concerns/forms/work_form.rb
@@ -15,7 +15,7 @@ module CurationConcerns
                     :representative_id, :thumbnail_id, :files,
                     :visibility_during_embargo, :embargo_release_date, :visibility_after_embargo,
                     :visibility_during_lease, :lease_expiration_date, :visibility_after_lease,
-                    :visibility, :ordered_member_ids]
+                    :visibility, :ordered_member_ids, :source]
 
       self.required_fields = [:title]
 
@@ -40,17 +40,10 @@ module CurationConcerns
 
       class << self
         # This determines whether the allowed parameters are single or multiple.
-        # By default it delegates to the model, but we need to override for
-        # 'rights' which only has a single value on the form.
+        # By default it delegates to the model.
         def multiple?(term)
-          case term.to_s
-          when 'rights'
-            false
-          when 'ordered_member_ids'
-            true
-          else
-            super
-          end
+          return true if term.to_s == 'ordered_member_ids'
+          super
         end
 
         # Overriden to cast 'rights' to an array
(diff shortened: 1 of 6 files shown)


app/forms/curation_concerns/forms/work_form.rb
```diff
@@ -44,8 +44,6 @@ class << self
         # 'rights' which only has a single value on the form.
         def multiple?(term)
```
mjgiarlo: Now that this method is only special-casing `'ordered_member_ids'`, would you consider refactoring this method to use the guard clause pattern?

```ruby
def multiple?(term)
  return true if term.to_s == 'ordered_member_ids'
  super
end
```

app/forms/curation_concerns/forms/work_form.rb
```diff
@@ -44,8 +44,6 @@ class << self
         # 'rights' which only has a single value on the form.
```
mjgiarlo: Can you strike `, but we need to override for 'rights' which only has a single value on the form`?

mjgiarlo:
Should we also add a view or presenter spec to cover the change made to the presenter? (I should expect to see source presented.)

app/models/concerns/curation_concerns/solr_document_behavior.rb
```diff
@@ -123,6 +123,10 @@ def read_groups
       fetch(Hydra.config.permissions.read.group, [])
     end
 
+    def source
+      self[Solrizer.solr_name('source')]
```
mjgiarlo: Should this method return `nil` or `[]` when there are no 'source' keys? I wonder if the `fetch()` pattern from above might be a better fit here.

eliotjordan: Yep. Imma fix rights too while I'm at it.

PR #836, opened 2016-06-03
Raise an error if no file_set_ids or member_ids were passed

jcoyne:
If we don't trap for this error state an infinite loop can arise.


tpendragon:
Spec?

jcoyne:
@tpendragon yeah, you didn't create a spec when you created that class.  Bad! @tpendragon 😧

jcoyne:
It appears this path is used when adding a child work.

PR #838, opened 2016-06-07
Mark the title field as required for the collection edit form

justcolin:
Fixes projecthydra/sufia#2106

The title field isn't being marked as required for collections in CC/Sufia.

Ongoing work: I am going to look into automatically getting this data from the model.

### DIFF
--- a/app/forms/curation_concerns/forms/collection_edit_form.rb
+++ b/app/forms/curation_concerns/forms/collection_edit_form.rb
@@ -11,6 +11,8 @@ module CurationConcerns
                     :representative_id, :thumbnail_id, :identifier, :based_near,
                     :related_url, :visibility]
 
+      self.required_fields = [:title]
+
       # Test to see if the given field is required
       # @param [Symbol] key a field
       # @return [Boolean] is it required or not
--- a/spec/features/collection_spec.rb
+++ b/spec/features/collection_spec.rb
@@ -17,6 +17,17 @@ feature 'collection' do
     Collection.destroy_all
   end
 
+  describe 'collection creation fields' do
+    before do
+      sign_in user
+      visit '/collections/new'
+    end
+
+    it 'visibly marks the title field as required' do
+      expect(page).to have_content 'Title required'
+    end
+  end
+
   describe 'create collection' do
     before do
       sign_in user


mjgiarlo:
@justcolin Can you add a test here that will break if your change is reverted? https://github.com/projecthydra/curation_concerns/blob/master/spec/forms/collection_edit_form_spec.rb

justcolin:
@mjgiarlo: If I do that I assume that I should also test the same thing for works, since they also have required fields that need this set in order to be marked.

mjgiarlo:
@justcolin You may if you're inclined, but I don't see that as a requirement for _this_ PR. (If you wanted to write that up as a separate ticket, that'd be cool too.)

spec/forms/collection_edit_form_spec.rb
```diff
@@ -67,4 +68,23 @@
...
+    let(:user) do
+      User.new(user_attributes) { |u| u.save(validate: false) }
+    end
+
+    before do
+      login_as user
+    end
+
+    it 'visibly marks the title field as required' do
+      visit '/concern/generic_works/new'
```
mjgiarlo: Shouldn't this be testing the collection edit path?

justcolin: I thought the edit and new forms are the same for collections.

spec/forms/collection_edit_form_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'spec_helper'
+include Warden::Test::Helpers
```
jcoyne: This is a form unit test, you should not have a request context of any sort.

spec/forms/collection_edit_form_spec.rb
mjgiarlo: But that's for `generic_works`, not collections, right? And yeah, new/edit forms are the same.

spec/forms/collection_edit_form_spec.rb
```diff
@@ -67,4 +68,23 @@
...
+      User.new(user_attributes) { |u| u.save(validate: false) }
+    end
+
+    before do
+      login_as user
+    end
+
+    it 'visibly marks the title field as required' do
+      visit '/concern/generic_works/new'
+      expect(page).to have_content('Title required')
```
jcoyne: This sort of thing should be done in a feature test `spec/features/*`, not in a unit test.

spec/forms/collection_edit_form_spec.rb
justcolin: Sorry, that was a typo. Pushing a fix shortly.

spec/forms/collection_edit_form_spec.rb
justcolin: Without that line it said that it couldn't find the method `login_as`. Is there something else I should include to make that work?

spec/forms/collection_edit_form_spec.rb
justcolin: Would `spec/features/collection_spec.rb` make sense?

jcoyne: Yes, that would be perfect.

justcolin: Thanks!

PR #846, opened 2016-06-10
enables AddToWorkActor to remove all relationships to work_ids

### LINKED ISSUE #845: AddToWorkActor unable to remove all related in_works_ids
Passing an empty array to the update method returns true before performing the synchronization, which should just delete all of the existing relationships.

revgum:
Fixes #845

### DIFF
--- a/spec/actors/curation_concerns/work_actor_spec.rb
+++ b/spec/actors/curation_concerns/work_actor_spec.rb
@@ -204,6 +204,43 @@ describe CurationConcerns::Actors::GenericWorkActor do
         expect(old_parent.reload.members).to eq []
       end
     end
+    context 'without in_works_ids' do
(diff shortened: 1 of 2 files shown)


spec/actors/curation_concerns/work_actor_spec.rb
```diff
@@ -204,6 +204,35 @@
...
+        expect(old_parent.reload.members).to eq []
+      end
+    end
+    context 'with nil in_works_ids' do
+      let(:attributes) do
+        FactoryGirl.attributes_for(:generic_work).merge(
+          in_works_ids: nil
+        )
+      end
+      it "removes the old parent" do
```
jcoyne: Shouldn't it "do nothing?"

jcoyne:
@tpendragon does this look good to you?

PR #851, opened 2016-06-15
Improved content negotiation

jcoyne:

Encode suburis as hashcodes. Encode subject URIs separately (previously they would go to the controller for the main model).


tpendragon:
If you encode contained resources as hash-URIs, what happens to actual hash-uris of those contained resources? #list_source#1? I don't see any code to actually pull in those nodes, so is there a different strategy I'm missing?

jcoyne:
@tpendragon Good point. I haven't handled those yet.  Mostly this provides a framework for that work to occur.

jcoyne:
@tpendragon I fixed the hash code issue.

tpendragon:
Test failed.

jcoyne:
@tpendragon it's 💚 now.

PR #852, opened 2016-06-17
Change Edit path for Collections

njaffer:
Fixes this error:
screen shot 2016-06-17 at 2 25 15 pm

* update edit path for collection
* 
* 

### DIFF
--- a/spec/views/catalog/_action_menu_partials/_collection.html.erb_spec.rb
+++ b/spec/views/catalog/_action_menu_partials/_collection.html.erb_spec.rb
@@ -1,0 +1,14 @@
+require 'spec_helper'
+
+describe 'catalog/_action_menu_partials/_collection.erb' do
+  let(:document) { double }
+  subject { rendered }
+
+  context "when all the paths are valid" do
+    before do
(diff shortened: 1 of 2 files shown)


awead:
Would you be willing to write a short test that verifies this? You should be able to crib from https://github.com/projecthydra/curation_concerns/blob/master/spec/views/catalog/_action_menu_partials/_default.html.erb_spec.rb

awead:
I'm also willing to forgo the test, too 😄

njaffer:
added the test..

jcoyne:
I think the problem here is that the `document` does not have an id. So it's trying `edit_collections_path` (note the plural) and not `edit_collection_path` as it should.

spec/views/catalog/_action_menu_partials/_collection.html.erb_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe 'catalog/_action_menu_partials/_collection.erb' do
+  let(:document) { double }
```
jcoyne: It can't possibly draw the correct route without an identifier.

PR #854, opened 2016-06-20
Update Collection when a member work is deleted

### LINKED ISSUE #853: Total Items in Collection not updated when a work is deleted
### Descriptive summary
This is more obvious in Sufia, the total items on the collection page shows 1, even though the collection has no works. When the work is deleted, and it is part of a collection, should the option for delete work be disabled or give a notification that it is part of a collection?

### Related work
https://github.com/projecthydra/sufia/issues/2223

Link to related tickets or prior related work here.

njaffer:
Fixes #853 

Present tense short summary (50 characters or less)
When a work is deleted, check if the work id is a member id of a collection. If found, delete the member id and update the collection

* Add a method "remove_from_collection" in curation_concern_controller
* 
* 

### DIFF
--- a/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
+++ b/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
@@ -93,8 +93,17 @@ module CurationConcerns::CurationConcernController
     end
   end
 
+  def remove_from_collection
+    curation_concern.in_collection_ids.each do |id|
+      destination_collection = ::Collection.find(id)
+      destination_collection.members.delete(curation_concern)
+      destination_collection.update_index
+    end
+  end
+
   def destroy
     title = curation_concern.to_s
+    remove_from_collection
     curation_concern.destroy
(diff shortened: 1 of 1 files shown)


jcoyne:
I think this belongs in a DestroyActor. As you might remove an object in a batch (non HTTP) context

njaffer:
@jcoyne: Where is the DestroyActor defined?

jcoyne:
@njaffer It isn't yet. Can you make it?

njaffer:
@jcoyne: Ok, I will give it a try. Would it be similar to AddToCollectionActor? RemoveFromCollectionActor ? or should it be DestroyWorkActor?

jcoyne:
Please add a test to this code too.

jcoyne:
You could pattern it after AddToCollectionActor. But in this case you are deleting a work instead of updating or creating it.

njaffer:
@jcoyne: I will close this PR, I have opened another one with the actor. Please review it and let me know if that's the right place for that actor.

PR #859, opened 2016-06-22
Adds a method which allows for the assignment of children to parents.

### LINKED ISSUE #857: Add child to parent work
Currently in CC, a user has the ability to add parent works to children works. The ability to add a Child work to a parent work would be quite useful. 

Use case: As a user, I would like to be able to create a new work, then create a second new work and assign the first work created as a child to the second work.

straleyb:
Fixes #857 

This update achieves the ability to add new children to a parent work. We added a new method in the same module where it orders the children in the context of a parent work. This method breaks any relations the parent has to children and rebuilds those relationships based on the ordered_member_ids that are passed through the form.

### DIFF
--- a/app/actors/curation_concerns/actors/apply_order_actor.rb
+++ b/app/actors/curation_concerns/actors/apply_order_actor.rb
@@ -3,10 +3,26 @@ module CurationConcerns
     class ApplyOrderActor < AbstractActor
       def update(attributes)
         ordered_member_ids = attributes.delete(:ordered_member_ids)
+        sync_members(ordered_member_ids)
         apply_order(ordered_member_ids) && next_actor.update(attributes)
       end
 
       private
+
+        def sync_members(ordered_member_ids)
+          return true if ordered_member_ids.nil?
+          existing_members_ids = curation_concern.member_ids
+          (existing_members_ids - ordered_member_ids).each do |old_id|
+            work = ::ActiveFedora::Base.find(old_id)
+            curation_concern.ordered_members.delete(work)
+          end
+
+          (ordered_member_ids - existing_members_ids).each do |work_id|
+            work = ::ActiveFedora::Base.find(work_id)
+            curation_concern.ordered_members << work
+          end
+          true
+        end
 
         def apply_order(new_order)
           return true unless new_order
--- a/spec/actors/curation_concerns/apply_order_actor_spec.rb
+++ b/spec/actors/curation_concerns/apply_order_actor_spec.rb
@@ -1,0 +1,60 @@
+require 'spec_helper'
+
+describe CurationConcerns::Actors::ApplyOrderActor do
+  describe '#update' do
+    let(:curation_concern) { create(:work_with_two_children, user: user) }
+
+    let(:user) { create(:admin) }
+
+    subject do
+      CurationConcerns::Actors::ActorStack.new(curation_concern,
+                                               user,
+                                               [described_class,
+                                                CurationConcerns::Actors::GenericWorkActor])
+    end
+
+    context 'with ordered_member_ids that are already associated with the parent' do
+      let(:attributes) { { ordered_member_ids: ["Blah"] } }
+      let(:root_actor) { double }
+      before do
+        allow(CurationConcerns::Actors::RootActor).to receive(:new).and_return(root_actor)
+        allow(root_actor).to receive(:update).with({}).and_return(true)
+        curation_concern.apply_depositor_metadata(user.user_key)
+        curation_concern.save!
+      end
+      it "attaches the parent" do
+        expect(subject.update(attributes)).to be true
+      end
+    end
+  end
+
+  describe '#update' do
+    let(:user) { create(:admin) }
+    let(:curation_concern) { create(:work_with_one_child, user: user) }
+    let(:child) { GenericWork.new("Blah3") }
+
+    subject do
+      CurationConcerns::Actors::ActorStack.new(curation_concern,
+                                               user,
+                                               [described_class,
+                                                CurationConcerns::Actors::GenericWorkActor])
+    end
+
+    context 'with ordered_members_ids that arent associated with the curation concern yet.' do
+      let(:attributes) { { ordered_member_ids: ["Blah3"] } }
+      let(:root_actor) { double }
+      before do
+        allow(CurationConcerns::Actors::RootActor).to receive(:new).and_return(root_actor)
+        allow(root_actor).to receive(:update).with({}).and_return(true)
+        child.title = ["Generic Title"]
+        child.apply_depositor_metadata(user.user_key)
+        child.save!
+        curation_concern.apply_depositor_metadata(user.user_key)
+        curation_concern.save!
+      end
+      it "attaches the parent" do
+        expect(subject.update(attributes)).to be true
+      end
+    end
+  end
+end
--- a/spec/factories/generic_works.rb
+++ b/spec/factories/generic_works.rb
@@ -23,6 +23,19 @@ FactoryGirl.define do
       end
     end
 
+    factory :work_with_one_child do
+      before(:create) do |work, evaluator|
+        work.ordered_members << FactoryGirl.create(:generic_work, user: evaluator.user, title: ['A Contained Work'])
+      end
+    end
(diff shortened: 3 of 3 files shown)


app/actors/curation_concerns/actors/apply_order_actor.rb
```diff
@@ -3,11 +3,28 @@ module Actors
...
         ordered_member_ids = attributes.delete(:ordered_member_ids)
+        sync_members(ordered_member_ids)
         apply_order(ordered_member_ids) && next_actor.update(attributes)
       end
 
       private
 
+        def sync_members(ordered_member_ids)
+          return true if ordered_member_ids.nil?
+          existing_members_ids = curation_concern.members.select { |k| k.class.to_s != 'FileSet' }.map(&:id)
```
tpendragon: Why not FileSets? They're members too.

Also, `curation-concerns.members` is a really expensive operation. Can you just check the ordered_member_ids?

app/actors/curation_concerns/actors/apply_order_actor.rb
```diff
@@ -3,11 +3,28 @@ module Actors
...
+          (existing_members_ids - ordered_member_ids).each do |old_id|
+            work = ::ActiveFedora::Base.find(old_id)
+            curation_concern.ordered_members.delete(work)
+            curation_concern.save
```
tpendragon: This is a lot of saves.

app/actors/curation_concerns/actors/apply_order_actor.rb
```diff
@@ -3,11 +3,28 @@ module Actors
     class ApplyOrderActor < AbstractActor
       def update(attributes)
         ordered_member_ids = attributes.delete(:ordered_member_ids)
+        sync_members(ordered_member_ids)
         apply_order(ordered_member_ids) && next_actor.update(attributes)
       end
 
       private
 
+        def sync_members(ordered_member_ids)
```
tpendragon: Can we get a spec?

app/actors/curation_concerns/actors/apply_order_actor.rb
straleyb: I felt like filesets were unnecessary to iterate through when assigning children to parent works. Eventhough they are members, they arent considered "children", so we didn't want them in the list of existing_member_ids

app/actors/curation_concerns/actors/apply_order_actor.rb
straleyb: Do you think we could extract it from the loop and just save once after it finishes iterating through the id's?

app/actors/curation_concerns/actors/apply_order_actor.rb
straleyb: I didnt see a spec for ApplyOrderActor. Did I miss it somewhere? Or would the spec belong in the GenericWorksControllerSpec#update?

app/actors/curation_concerns/actors/apply_order_actor.rb
tpendragon: You're not really getting any speed benefits by skipping over them, comparing the ids is probably faster.

app/actors/curation_concerns/actors/apply_order_actor.rb
tpendragon: Yes.

app/actors/curation_concerns/actors/apply_order_actor.rb
tpendragon: Either works - an actor spec is probably better, if you can swing it. But https://coveralls.io/builds/6714170/source?filename=app%2Factors%2Fcuration_concerns%2Factors%2Fapply_order_actor.rb#L17 it's not covered.

app/actors/curation_concerns/actors/apply_order_actor.rb
revgum: So ordered_member_ids are passed in relative to the children that are being assigned to a work, and then we are trying to synchronize (add any new member ids that are included, and remove any existing member ids that weren't included).. FileSet's wouldn't be included ever from the front-end when dealing with children works. So we wanted to filter out and not compare FileSet's (because they would get removed).. this logic works pretty much the same way the AddWorkActor logic works. Should we chat to clarify?

tpendragon: I'm concerned about the ordering that would happen with that strategy. The order of FileSet members relative to child work members is important.

revgum: I feel like FileSet and Child Work Members, definitely in Sufia anyway, are two different things shoved into one "ordered_members". I don't have a well informed understanding of the implications or an approach to keep them separate..

tpendragon: That's a fair point, and it's looking like they'll be separated in PCDM 2.0, but for the moment a FileSet is a PCDM:Object, and so is any given work, so they're all collected as members.

app/actors/curation_concerns/actors/apply_order_actor.rb
```diff
@@ -3,11 +3,29 @@ module Actors
...
+          return true if ordered_member_ids.nil?
+          existing_members_ids = curation_concern.member_ids
+          (existing_members_ids - ordered_member_ids).each do |old_id|
+            work = ::ActiveFedora::Base.find(old_id)
+            curation_concern.ordered_members.delete(work)
+          end
+          curation_concern.save
```
tpendragon: Does an integration spec pass if this actor just doesn't save anything? It didn't before, and it'd make things that much quicker.

straleyb: Yeah I can check it out. 👍

spec/actors/curation_concerns/apply_order_actor_spec.rb
```diff
@@ -50,17 +83,18 @@
...
       before do
         allow(CurationConcerns::Actors::RootActor).to receive(:new).and_return(root_actor)
         allow(root_actor).to receive(:update).with({}).and_return(true)
-        g = child
-        g.save
         curation_concern.apply_depositor_metadata(user.user_key)
         curation_concern.save!
       end
       it "attaches the parent" do
+        child.title = ["Generic Title"]
```
jcoyne: Why did this stuff move out of the `before` block.  It seems like that's the natural place for it.

jcoyne:
Please squash your commits and then 👍 from me.

spec/actors/curation_concerns/apply_order_actor_spec.rb
straleyb: I was experimenting with stuff. I'll extract thats stuff out and squash those commits and ill ping ya here. @jcoyne

straleyb:
Alright this is ready I think. @jcoyne @mjgiarlo

spec/actors/curation_concerns/apply_order_actor_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'spec_helper'
+
+describe CurationConcerns::Actors::ApplyOrderActor do
+  describe '#update' do
+    let(:curation_concern) { create(:work_with_two_children, user: user) }
+
+    let(:user) { create(:admin) }
+
+    before do
+      allow(described_class).to receive(:update).with(attributes)
```
tpendragon: `update` is an instance method, so I don't think this is doing anything. Delete?

spec/actors/curation_concerns/apply_order_actor_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+  end
+
+  describe '#update' do
+    let(:user) { create(:admin) }
+    let(:curation_concern) { create(:work_with_one_child, user: user) }
+    let(:child) { GenericWork.new("Blah3") }
+
+    before do
+      allow(described_class).to receive(:update).with(attributes)
```
tpendragon: Same as above.

tpendragon:
Couple notes about the test, then I'm :+1:

tpendragon:
Squash commits please.

PR #866, opened 2016-06-28
Reload derivatives to index extracted text

awead:

When text is extracted from a file, it is not available on the file set
until the data is reloaded from Fedora. Reloading it makes it available
for indexing. Fixes projecthydra/sufia#1813.

### DIFF
--- a/app/jobs/create_derivatives_job.rb
+++ b/app/jobs/create_derivatives_job.rb
@@ -8,7 +8,9 @@ class CreateDerivativesJob < ActiveJob::Base
     filename = CurationConcerns::WorkingDirectory.find_or_retrieve(file_id, file_set.id)
 
     file_set.create_derivatives(filename)
-    # The thumbnail is indexed in the solr document, so reindex
+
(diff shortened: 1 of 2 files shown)


tpendragon:
I'd prefer if the test ensured the text was indexed rather than #reload was called. Is that possible?

awead:
@tpendragon it's tested a bit more here:
https://github.com/projecthydra/curation_concerns/blob/master/spec/indexers/file_set_indexer_spec.rb#L80

Do you want an integration test that runs through the whole derivative creation process for a pdf?

tpendragon:
@awead At this point, I think so, unless it's very expensive. Do we not have that happening anywhere now?

awead:
@tpendragon not that I know of. I'll put one in and see how badly it slows down the suite.

awead:
@tpendragon try this. Didn't seem to slow things down much locally, but we'll see what Travis says.

mjgiarlo:
@awead what's the verdict?

PR #867, opened 2016-06-28
Remove Google analytics, ref #827

### LINKED ISSUE #827: Including Google Analytics id in config does not permit application to start.
### Descriptive summary

Include what version of CurationConcerns relates to this issue (0.7.x, 0.8.x, HEAD, etc.) if appropriate, and any relevant tracebacks if you're reporting a bug.

```
.../curation_concerns/.internal_test_app/config/initializers/curation_concerns.rb:20:in `block in <top (required)>': undefined method `google_analytics_id=' for #<CurationConcerns::Configuration:0x00560fcd9b0388> (NoMethodError)
	from .../curation_concerns/lib/curation_concerns/configuration.rb:10:in `configure'
	from .../curation_concerns/.internal_test_app/config/initializers/curation_concerns.rb:1:in `<top (required)>'

```

### Expected behavior
Google analytics id to be included in relevant pages via `app/views/shared/_ga.html.erb`

### Actual behavior
Application fails to start

1. Generate .internal_test_app
1. Modify `.internal_test_app/config/initializers/curation_concerns.rb` by uncommenting the `google_analytics_id=` line.
3. Run the application `bundle exec rails s`

awead:
Fixes #827 

Leave Google analytics up to the implementor, either Sufia or other.

### DIFF
--- a/app/views/layouts/boilerplate.html.erb
+++ b/app/views/layouts/boilerplate.html.erb
@@ -11,7 +11,6 @@
     <%= csrf_meta_tag %>
     <%= stylesheet_link_tag 'application' %>
     <%= javascript_include_tag 'application' %>
-    <%= render partial: 'shared/ga', formats: [:html] %>
   </head>
   <body>
 
--- a/app/views/layouts/error.html.erb
+++ b/app/views/layouts/error.html.erb
@@ -9,8 +9,6 @@
 
     <!-- application js -->
     <%= javascript_include_tag 'application' %>
-
-    <%= render partial: 'shared/ga', formats: [:html] %>
   </head>
 
 <body>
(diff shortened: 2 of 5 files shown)


jcoyne:
Not a fan. Especially since we have a lot of analytics reporting code that depends on this.

awead:
@jcoyne yeah, especially since we'll probably be moving them back together again, but I'm just responding to the issue.

escowles:
There was a lot of support for removing GA in #827.  IMHO, it would be good to remove the Google-specific code in `app/views/shared/_ga.html.erb` and leave that to implementing applications.

escowles:
A compromise would be to move the GA javascript into a helper that would only output the javascript if the GA id was configured.  This would address the original issue (missing config causing problems) and probably make it easier to completely disable the analytics, or swap in Piwik or something else.

jcoyne:
I'm not going to block the issue, but I think it should be here eventually, and I'd rather just fix it in CC instead of move it to Sufia and then move it back at some later time.

mjgiarlo:
I'm inclined to agree with @jcoyne, and I dig @escowles's compromise suggestion.

PR #869, opened 2016-07-01
Updates for Blacklight 6.3 compatibility

awead:
Changes required to support Blacklight 6.3.


awead:
I'm not completely happy with this solution because I think it needs to be fixed in Blacklight, but it unblocks things.

escowles:
That's good enough for me.

PR #872, opened 2016-07-07
Remove the BlacklightOverridesHelper. Fixes #871

### LINKED ISSUE #871: Remove BlacklightOverridesHelper
https://github.com/projecthydra/curation_concerns/blob/master/app/helpers/curation_concerns/blacklight_overrides_helper.rb

This helper is required because the CollectionsController uses the `catalog/index_default` partial from Blacklight.  This partial is intended to be used in the `index` action, so it calls `render_index_field_value`.  This method calls presenter with the current action name, which is `edit`, which is not a valid argument for Blacklight.  I think that copying the partial and instantiating the correct presenter would be more clear than this override.

jcoyne:


jcoyne:
The broken tests should be fixed by #873

awead:
LGTM. I thought I would see errors for edit actions, but if this is green, it's good for me.

PR #875, opened 2016-07-07
Allow a FileSet or an id for FileSetAuditService, fixes #874

### LINKED ISSUE #874: FileSetAuditService should only call Fedora when needed
### Descriptive summary

FileSetAuditService doesn't always need the complete Fedora object in order to do its job. We should allow an id for initialization, then instantiate the full object when required.

### Related work

https://github.com/projecthydra/sufia/issues/2326

awead:
Fixes #874 

Allows `logged_audit_status` to be called without invoking Fedora. Required for https://github.com/projecthydra/sufia/issues/2326

### DIFF
--- a/app/services/curation_concerns/file_set_audit_service.rb
+++ b/app/services/curation_concerns/file_set_audit_service.rb
@@ -1,8 +1,15 @@
 module CurationConcerns
   class FileSetAuditService
-    attr_reader :file_set
+    attr_reader :file_set, :id
+
+    # @param file_set [ActiveFedora::Base, String] file_set
     def initialize(file_set)
-      @file_set = file_set
+      if file_set.is_a?(String)
+        @id = file_set
+      else
+        @id = file_set.id
+        @file_set = file_set
+      end
     end
 
     NO_RUNS = 999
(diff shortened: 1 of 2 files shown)


jcoyne:
@awead can you amend the commit message to include the justification:

> Allows logged_audit_status to be called without invoking Fedora.

That's important and I don't want it to be lost.

escowles:
@jcoyne Can't you amend the commit message using the squash-and-merge feature?

jcoyne:
@escowles Isn't that only available if there is more than one commit?

jcoyne:
@escowles It looks like if I do that, we won't see a merge commit, so that would change the way we've traditionally done workflow.

escowles:
@jcoyne I'm pretty sure you can squash even if there's only 1 commit, but yes it would omit the merge commit.  I'm fine with having @awead update the commit with a new commit message.

Maybe we should bring up the squash-and-merge on the next Hydra tech call to see if we're OK using it?

awead:
@jcoyne no problem. @escowles thanks for the tip. I'll all about making commit messages more informative!

PR #878, opened 2016-07-11
Allow https uris

### LINKED ISSUE #877: Support uris using HTTPS
### Descriptive summary

This will fail if you're running Fedora over https:

https://github.com/projecthydra/curation_concerns/blob/master/app/services/curation_concerns/persist_directly_contained_output_file_service.rb#L17

I'm assuming this is not on purpose?

### Expected behavior

It should accept URIs with https.

### Actual behavior

It does not accept URIs using https.

awead:
Fixes #877 

I've run the test suite locally against a SSL-enabled Fedora, however, I had to disable certificate verification because it was a self-signed cert. But, everything was going over ssl and there were no other failures. I also ran Sufia's test suite agains this branch of CurationConcerns, and it worked fine too.

So, I'd say this probably all we need to do to support SSL.

### DIFF
--- a/app/services/curation_concerns/persist_directly_contained_output_file_service.rb
+++ b/app/services/curation_concerns/persist_directly_contained_output_file_service.rb
@@ -14,7 +14,7 @@ module CurationConcerns
       o_name = determine_original_name(file)
       m_type = determine_mime_type(file)
       uri = URI(directives.fetch(:url))
-      raise ArgumentError, "#{uri} is not an http uri" unless uri.scheme == 'http'
+      raise ArgumentError, "#{uri} is not an http(s) uri" unless uri.is_a?(URI::HTTP)
       file_set = ActiveFedora::Base.find(ActiveFedora::Base.uri_to_id(uri.to_s))
       remote_file = file_set.send("build_#{directives.fetch(:container)}".to_sym)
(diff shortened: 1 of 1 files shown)


awead:
If anyone else @escowles @jpstroop @mjgiarlo @jcoyne wants to smoke-test this a bit more, that'd be great, but otherwise this seems to do the trick.

app/services/curation_concerns/persist_directly_contained_output_file_service.rb
```diff
@@ -14,7 +14,7 @@ def self.call(stream, directives)
       o_name = determine_original_name(file)
       m_type = determine_mime_type(file)
       uri = URI(directives.fetch(:url))
-      raise ArgumentError, "#{uri} is not an http uri" unless uri.scheme == 'http'
+      raise ArgumentError, "#{uri} is not an http(s) uri" unless uri.scheme =~ /http|https/
```
cbeer: What do you think about `uri.is_a? URI::HTTP`?

```ruby
> URI.parse('https://asdf').is_a? URI::HTTP
=> true
```

awead: Works for me!

PR #879, opened 2016-07-13
Use ordered member ids instead of ids

straleyb:
We had to change from using #member_ids to using #ordered_member_ids. member_ids was returning the full list of members including the one we were adding as a child in the attributes. This prevented the relationship from being persisted.


jcoyne:
@tpendragon can you review this please?

PR #883, opened 2016-07-22
Single Use Links UI

### LINKED ISSUE #278: Re-enable single-use links in the UI
Although all of the single-use link functionality is in CurationConcerns, there isn't a way actually generate one from the interface. Single-use links are no longer exposed in the UI.

1. Single-use links will be exposed in the UI for `FileSets` only (not for e.g., `GenericWorks` or other curation concerns).
1. Only one single-use link will be generated for each `FileSet`, allowing download access to the `original_file` within the `FileSet`.
1. Clicking the single-use link button for a `FileSet` should not redirect to a new view, but rather it should pop up a (Bootstrap) modal showing the single-use link, highlighted, with a GitHub-like "copy me" button beside it. Consider using https://clipboardjs.com/ for this.

You may want to review and adapt/re-use the following bits of implementation from Sufia:

* https://github.com/projecthydra/sufia/blob/af6a75095047ae3cd8aeffd145713305aa63f041/app/assets/javascripts/sufia/single_use_link.js
* https://github.com/projecthydra/sufia/blob/af6a75095047ae3cd8aeffd145713305aa63f041/app/assets/javascripts/sufia.js#L39
* https://github.com/projecthydra/sufia/blob/af6a75095047ae3cd8aeffd145713305aa63f041/spec/javascripts/single_use_link_spec.js.coffee
* https://github.com/projecthydra/sufia/blob/af6a75095047ae3cd8aeffd145713305aa63f041/app/views/my/_action_menu.html.erb#L8-L11

This note may be useful or it may be outdated:

> Make sure the link renders if a title isn't set on the file. This [link](https://github.com/projecthydra-labs/curation_concerns/blob/39f7b89b43c8add7ea71547362663c69005931ba/app/views/curation_concerns/single_use_links/new_download.html.erb#L4) and this [link](https://github.com/projecthydra-labs/curation_concerns/blob/39f7b89b43c8add7ea71547362663c69005931ba/app/views/curation_concerns/single_use_links/new_show.html.erb#L4) don't render when the title attribute isn't set.

awead:
Fixes #278 

Here's the preliminary idea.

sul

Need some feedback at this point before it can be finished.

### DIFF
--- a/app/controllers/concerns/curation_concerns/single_use_links_controller_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/single_use_links_controller_behavior.rb
@@ -25,6 +25,16 @@ module CurationConcerns
       render text: curation_concerns.show_single_use_link_url(@su.downloadKey)
     end
 
+    def index
+      links = SingleUseLink.where(itemId: params[:id])
+      render partial: 'curation_concerns/file_sets/single_use_link_rows', locals: { single_use_links: links }
+    end
+
+    def destroy
+      SingleUseLink.find_by_downloadKey(params[:link_id]).destroy
+      head :ok
+    end
+
     protected
 
       def authorize_user!
--- a/app/presenters/curation_concerns/file_set_presenter.rb
+++ b/app/presenters/curation_concerns/file_set_presenter.rb
@@ -33,5 +33,9 @@ module CurationConcerns
     def link_name
       current_ability.can?(:read, id) ? label : 'File'
     end
+
+    def single_use_links
+      @single_use_links ||= SingleUseLink.where(itemId: id)
+    end
   end
 end
--- a/app/views/curation_concerns/file_sets/_single_use_link_rows.html.erb
+++ b/app/views/curation_concerns/file_sets/_single_use_link_rows.html.erb
@@ -1,0 +1,18 @@
+<% single_use_links.each_with_index do |link, index| %>
+  <% type = link.path =~ /downloads/ ? 'Download' : 'Show' %>
+  <% url_helper = link.path =~ /downloads/ ? 'download_single_use_link_url' : 'show_single_use_link_url' %>
+  <tr>
+    <td><%= type %></td>
+    <td><%= link.downloadKey %></td>
+    <td><%= link.expires %></td>
+    <td>
+      <button class="btn btn-xs btn-default copy-single-use-link"
+              data-clipboard-text="<%= curation_concerns.send(url_helper, link.downloadKey) %>"
+              data-toggle="tooltip" data-placement="bottom" title="Copied!">
+        Copy to Clipboard
+      </button>
+      <%= link_to "Delete", curation_concerns.delete_single_use_link_path(params[:id], link),
+          class: 'btn btn-xs btn-danger delete-single-use-link' %>
+    </td>
+  </tr>
+<% end %>
--- a/app/views/curation_concerns/file_sets/_single_use_links.html.erb
+++ b/app/views/curation_concerns/file_sets/_single_use_links.html.erb
@@ -1,0 +1,20 @@
+<table class="table table-striped <%= dom_class(presenter) %> single-use-links">
+  <caption class="table-heading"><h2>Single-Use Links</h2></caption>
+  <thead>
+    <tr><th>Link</th><th>Key</th><th>Expires</th><th>Actions</th></tr>
+  </thead>
+  <tbody data-url="<%= curation_concerns.generated_single_use_links_path(presenter) %>">
+    <% if presenter.single_use_links.empty? %>
+      <tr><td>No links have been generated</td></tr>
+    <% else %>
+      <%= render 'single_use_link_rows', single_use_links: presenter.single_use_links %>
+    <% end %>
+  </tbody>
+</table>
+
+<div class="form_actions">
+  <%= link_to "Generate Download Link", curation_concerns.generate_download_single_use_link_path(presenter),
+      class: 'btn btn-default generate-single-use-link' %>
+  <%= link_to "Generate Show Link", curation_concerns.generate_show_single_use_link_path(presenter),
+      class: 'btn btn-default generate-single-use-link' %>
+</div>
--- a/spec/factories/single_use_links.rb
+++ b/spec/factories/single_use_links.rb
@@ -1,0 +1,13 @@
+FactoryGirl.define do
+  factory :single_use_link do
+    factory :show_link do
+      itemId 'fs-id'
+      path '/concerns/generic_work/1234'
+    end
+
+    factory :download_link do
+      itemId 'fs-id'
+      path '/downloads/1234'
+    end
+  end
+end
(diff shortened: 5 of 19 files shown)


mjgiarlo:
The last few times we discussed this, I believe we concluded that show links were more trouble than they were worth because they display a site full of links, only some of which will work for the user. It's an awkward user experience. Before I go over the changes in full, I'd lean towards excising that change but take that with a grain of salt. I'll review this fully later today when I'm done with my meetings.

app/controllers/concerns/curation_concerns/single_use_links_controller_behavior.rb
```diff
@@ -25,6 +25,16 @@ def create_show
       render text: curation_concerns.show_single_use_link_url(@su.downloadKey)
     end
 
+    def index
+      links = SingleUseLink.where(itemId: params[:id])
```
mjgiarlo: `s/itemId/item_id/`

app/controllers/concerns/curation_concerns/single_use_links_controller_behavior.rb
```diff
@@ -25,6 +25,16 @@ def create_show
...
+      render partial: 'curation_concerns/file_sets/single_use_link_rows', locals: { single_use_links: links }
```
mjgiarlo: I believe this can be simplified to:

```ruby
render 'curation_concerns/file_sets/single_use_link_rows', single_use_links: links
```

app/presenters/curation_concerns/file_set_presenter.rb
```diff
@@ -33,5 +33,9 @@ def page_title
     def link_name
       current_ability.can?(:read, id) ? label : 'File'
     end
+
+    def single_use_links
+      @single_use_links ||= SingleUseLink.where(itemId: id)
```
mjgiarlo: `s/itemId/item_id/`

app/views/curation_concerns/file_sets/_single_use_link_rows.html.erb
```diff
@@ -0,0 +1,18 @@
+<% single_use_links.each_with_index do |link, index| %>
+  <% type = link.path =~ /downloads/ ? 'Download' : 'Show' %>
+  <% url_helper = link.path =~ /downloads/ ? 'download_single_use_link_url' : 'show_single_use_link_url' %>
+  <tr>
+    <td><%= type %></td>
+    <td><%= link.downloadKey %></td>
```
mjgiarlo: I'm `+0` on displaying the download key to users. Not sure it's meaningful for many of them. Certainly doesn't *hurt* to show it, so I won't complain too much about this.

app/views/curation_concerns/file_sets/_single_use_links.html.erb
```diff
@@ -0,0 +1,20 @@
+<table class="table table-striped <%= dom_class(presenter) %> single-use-links">
+  <caption class="table-heading"><h2>Single-Use Links</h2></caption>
+  <thead>
+    <tr><th>Link</th><th>Key</th><th>Expires</th><th>Actions</th></tr>
+  </thead>
+  <tbody data-url="<%= curation_concerns.generated_single_use_links_path(presenter) %>">
+    <% if presenter.single_use_links.empty? %>
+      <tr><td>No links have been genereated</td></tr>
```
mjgiarlo: `s/genereated/generated/`

app/views/curation_concerns/file_sets/_single_use_links.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <% else %>
+      <%= render partial: 'single_use_link_rows', locals: { single_use_links: presenter.single_use_links } %>
```
mjgiarlo: Can be simplified to:

```ruby
render 'single_use_link_rows', single_use_links: presenter.single_use_links 
```

spec/factories/single_use_links.rb
```diff
@@ -0,0 +1,13 @@
+FactoryGirl.define do
+  factory :single_use_link do
+    factory :show_link do
+      itemId 'fs-id'
```
mjgiarlo: `s/itemId/item_id/`

spec/factories/single_use_links.rb
```diff
@@ -0,0 +1,13 @@
...
+      path '/concerns/generic_work/1234'
+    end
+
+    factory :download_link do
+      itemId 'fs-id'
```
mjgiarlo: `s/itemId/item_id/`

spec/presenters/curation_concerns/file_set_presenter_spec.rb
```diff
@@ -77,4 +77,11 @@
       it { is_expected.to eq 'File' }
     end
   end
+
+  describe "single_use_links" do
+    it "returns ActiveRecord::Relation of all single use links for the file set" do
+      expect(SingleUseLink).to receive(:where).with(itemId: presenter.id)
```
mjgiarlo: `s/itemId/item_id/`

awead: `itemId` is coming from the model https://github.com/projecthydra/curation_concerns/blob/master/app/models/single_use_link.rb#L8

mjgiarlo: Ah, yes, thanks. I'll retract those comments.

app/controllers/concerns/curation_concerns/single_use_links_controller_behavior.rb
awead: @mjgiarlo you get `ActionView::MissingTemplate - Missing template curation_concerns/file_sets/single_use_link_rows` because it's expecting it to be a view template and not a partial since we'er in the `index` action.

mjgiarlo: I stand corrected. Context is everything. :) Thanks, @awead.

app/views/curation_concerns/file_sets/_single_use_links.html.erb
awead: this works.

app/views/curation_concerns/file_sets/_single_use_link_rows.html.erb
awead: It exposes everything, but the distinction isn't really explained. If we only want to show one kind of link, I would argue that it should be the "show" link, although we don't have to name it as such. In the process of visiting a show link, a download link is created automatically. However, if the user visited the show link, but never actually downloaded the content, the auto-created download link would still be hanging around. The user might want to be able to delete it, if they felt the need to control access, but no one would know the key of that download link anyway, so it's probably a moot issue.

awead:
@mjgiarlo I've addressed your comments, and have redone the JS in ECMA6 form, with some tests. It's not super-covered, but it's better than the previous non-existent coverage!

mjgiarlo:
As soon as Travis is :green_apple:, I'll be happy to merge it, show links and all.

awead:
@mjgiarlo this is ready. I added one more commit for engine cart, but it wasn't impacting the build.

awead:
@mjgiarlo this needs some more work to get the Js code working with Sufia.

mjgiarlo:
@awead OK. Thanks for testing Sufia integration before I merged this.

awead:
@mjgiarlo this is ready to go now, and I have a follow-up PR in Sufia that makes use of these changes.

mjgiarlo:
👏

PR #885, opened 2016-08-02
Add SingleUseLinkPresenter with internationalized settings

awead:
Addresses issues raised in projecthydra/sufia#2383

Adds a SingleUseLinkPresenter to encapsulate presentation logic for links, as well as internationalization settings.

### DIFF
--- a/app/presenters/curation_concerns/single_use_link_presenter.rb
+++ b/app/presenters/curation_concerns/single_use_link_presenter.rb
@@ -1,0 +1,52 @@
+module CurationConcerns
+  class SingleUseLinkPresenter
+    include ActionView::Helpers::TranslationHelper
+
+    attr_reader :link
+
+    delegate :downloadKey, :expired?, :to_param, to: :link
+
+    # @param link [SingleUseLink]
+    def initialize(link)
+      @link = link
+    end
+
+    def human_readable_expiration
+      if hours < 1
+        t('curation_concerns.single_use_links.expiration.lesser_time')
+      else
+        t('curation_concerns.single_use_links.expiration.time', value: hours)
+      end
+    end
+
+    def short_key
+      link.downloadKey.first(6)
+    end
+
+    def link_type
+      if download?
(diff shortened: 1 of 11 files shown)


app/presenters/curation_concerns/single_use_link_presenter.rb
```diff
@@ -0,0 +1,35 @@
...
+      t('curation_concerns.single_use_links.expiration.time', value: hours)
+    end
+
+    def short_key
+      link.downloadKey.first(6)
+    end
+
+    def define_link
+      if link.path =~ /downloads/
+        @link_type = t('curation_concerns.single_use_links.download.type')
```
mjgiarlo: Why not declare `link_type` and `url_helper` as methods in the presenter so we don't have to inject more ivars?

awead: so I don't have to duplicate the `if link.path =~ /downloads/` logic.

mjgiarlo: @awead Extract that to its own private method, like:

```ruby
def download_link?
  link.path =~ /downloads/
  # or something that returns true/false instead of truthy/falsey like link.path.include? 'downloads' ?
end
```

I'd rather see duplication of a conditional than more ivars.

awead: Ok, but I don't understand why ivars are viewed as a negative. Is there a performance impact?

mjgiarlo: No performance impact that I'm aware of. It's more in line with the lessons we learned about OOP design from Sandi Metz, namely that it's preferable to send a message than to rely on an ivar, and it's also easier to stub and override methods than it is to deal with ivars. It's also one of the "Sandi rules" to make it so that a controller only has to know about one ivar, which in our world is typically a presenter instance.

awead:
@mjgiarlo ready.

PR #888, opened 2016-08-09
[WIP] Adding support for member_of_collections links from Works to Collections

escowles:
To address performance problems with large numbers of links to other Fedora objects, support linking from Objects to Collections they are members of.  This supports the typical use case where each Object is a member of a small number of Collections.  But there are other use cases (e.g., popular objects in a large number of user collections) where the existing Collection->Object link may perform better.

**Question:** Should we keep both kinds of collection membership?
    If so, should we develop different terminology to keep the membership lists separate?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,11 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in curation_concerns.gemspec
 gemspec
 
+gem 'hydra-works', github: 'projecthydra/hydra-works', branch: 'member_of'
+gem 'active-fedora', github:'projecthydra/active_fedora', branch: 'master'
+gem 'hydra-derivatives', github:'projecthydra/hydra-derivatives', branch: 'master'
+gem 'hydra-pcdm', github:'projecthydra/hydra-pcdm', branch: 'member_of'
+
 group :development, :test do
   gem 'simplecov', '~> 0.9', require: false
   gem 'coveralls', require: false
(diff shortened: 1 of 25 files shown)


mjgiarlo:
@escowles shall we discuss this in tomorrow's Hydra Tech call?

escowles:
@mjgiarlo Yes, wanted to put this up for discussion — just added it to tomorrow's Hydra Tech Call agenda: https://wiki.duraspace.org/display/hydra/Hydra+Tech+Call+2016-08-10

Gemfile
```diff
@@ -3,6 +3,11 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in curation_concerns.gemspec
 gemspec
 
+gem 'hydra-works', github: 'projecthydra/hydra-works', branch: 'member_of'
+gem 'active-fedora', github:'projecthydra/active_fedora', branch: 'master'
```
jcoyne: I think `active-fedora` and `hydra-derivatives` can point at released versions now.

app/models/concerns/curation_concerns/solr_document_behavior.rb
```diff
@@ -23,7 +23,7 @@ def to_s
     #   link_to '...', SolrDocument(:id => 'bXXXXXX5').new => <a href="/dams_object/bXXXXXX5">...</a>
     def to_model
       @model ||= begin
-        m = ActiveFedora::Base.load_instance_from_solr(id, self)
+        m = ActiveFedora::Base.find(id)
```
jcoyne: 👏

no-reply:
This was discussed in [Hydra Tech 2016-08-10](https://wiki.duraspace.org/display/hydra/Hydra+Tech+Call+2016-08-10) (see point 3).

The decision was basically to punt this.

tpendragon:
@escowles Do we close this for #901 ?

escowles:
@tpendragon Yes, this branch is a proof-of-concept, but #901 is an actual working implementation that we're using in Plum.  So I'll close this branch in favor of that.

PR #889, opened 2016-08-09
Test with Rails 5

jcoyne:
Had to remove `rdf-vocab` as a dependency because bundler is complaining:

```
Your bundle requires gems that depend on each other, creating an infinite loop. Please remove either gem
'linkeddata' or gem 'rdf-vocab' and try again.
```


jcoyne:
I think this is blocked by #890

jcoyne:
Now blocked by https://github.com/projecthydra/active_fedora/issues/1122

PR #890, opened 2016-08-09
Remove skip of callback that doesn't exist

jcoyne:


cbeer:
Looks like it came from worthwhile? https://github.com/projecthydra-deprecated/worthwhile

PR #893, opened 2016-08-09
deletes member and persists changes

### LINKED ISSUE #892: ApplyOrderActor does not properly sync members
When removing members during a sync, calling `curation_concern.ordered_members.delete(work)` does not fully remove the relationship and causes a PCDM ancestor exception if that work is added as a parent (in_works) because the relationship exists. 

Observed this bug in testing Sufia WorksWithinWorks functionality;

1. Add an existing work as a child
2. Remove the work as a child
3. Add the work as a parent causes an exception because the member wasn't entirely removed in step 2

revgum:
fixes #892


mjgiarlo:
@revgum This could use tests, good sir.

revgum:
@mjgiarlo The lines I added are covered by other tests in the suite.. should I write something specific for removing the member?

mjgiarlo:
Yes, please. What we need are some tests that break in `master` but your code changes make pass.

PR #894, opened 2016-08-10
Remove use of deprecated method load_instance_from_solr

jcoyne:


app/controllers/concerns/curation_concerns/file_sets_controller_behavior.rb
```diff
@@ -67,7 +67,8 @@ def show
         wants.html { presenter }
         wants.json do
           # load and authorize @curation_concern manually because it's skipped for html
-          self.curation_concern ||= curation_concern_type.load_instance_from_solr(params[:id])
+          # TODO: is there any reason we couldn't use `presenter' here?
+          self.curation_concern ||= curation_concern_type.find(params[:id])
```
awead: Not that I know of. This is going to cause unwanted calls to Fedora.

app/controllers/concerns/curation_concerns/single_use_links_controller_behavior.rb
```diff
@@ -44,7 +44,7 @@ def authorize_user!
       end
 
       def asset
-        @asset ||= ActiveFedora::Base.load_instance_from_solr(params[:id])
+        @asset ||= ActiveFedora::Base.find(params[:id])
```
awead: I think I put this in here to get show pages running on all Solr and no Fedora. Can we replace it with a ActiveFedora::SolrService call instead?

awead: No, I didn't, but it would make show pages for FileSets reliant on Fedora and did fix that last month.

app/controllers/concerns/curation_concerns/single_use_links_controller_behavior.rb
```diff
@@ -40,11 +41,11 @@ def destroy
...
 
       def authorize_user!
-        authorize! :edit, asset
+        authorize! :edit, params[:id]
       end
 
-      def asset
-        @asset ||= ActiveFedora::Base.load_instance_from_solr(params[:id])
+      def asset_show_path
+        polymorphic_path([main_app, fetch(params[:id]).last])
```
escowles: Why the `last` call here? Can there be more than one id?

jcoyne: `#fetch()` returns a tuple, the first element is the solr response and the second is the solr document, which we want.

PR #895, opened 2016-08-10
Don't use the deprecated devise helpers

jcoyne:


curation_concerns.gemspec
```diff
@@ -19,6 +19,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency 'hydra-head', '>= 10.0.0', '< 11'
+  spec.add_dependency 'active-fedora', '>= 10.0.0', '< 11.a'
```
cbeer: What's the thinking behind this new dependency? The version constraints strike me as unusual.

jcoyne: Yeah, unusual: https://github.com/bundler/bundler/issues/4858#issuecomment-238554093

cbeer: Ew.

PR #900, opened 2016-08-16
Use a longer id for fixtures

jcoyne:

This avoids Fedora from showing errors like this:
```
ERROR 16:29:39.032 (InvalidResourceIdentifierExceptionMapper)
InvalidResourceIdentifierExceptionMapper caught an exception: Path
contains empty element! /test/Bl/ah//l/is/Blah/list_source

ERROR 16:29:39.053 (InvalidResourceIdentifierExceptionMapper)
InvalidResourceIdentifierExceptionMapper caught an exception: Path
contains empty element! /test/Bl/ah//l/is/Blah/list_source
```

### DIFF
--- a/spec/actors/curation_concerns/apply_order_actor_spec.rb
+++ b/spec/actors/curation_concerns/apply_order_actor_spec.rb
@@ -14,7 +14,7 @@ describe CurationConcerns::Actors::ApplyOrderActor do
     end
 
     context 'with ordered_member_ids that are already associated with the parent' do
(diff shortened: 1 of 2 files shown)


cbeer:
Does this need to be reported upstream, or is something in our stack generating that bad path in the first place?

jcoyne:
@cbeer I think this code https://github.com/projecthydra-labs/active_fedora-noid/blob/master/lib/active_fedora/noid.rb#L21 is expecting us to have ids with a certain number minimum number of characters.

PR #901, opened 2016-08-16
Reverse Collection membership

escowles:
Improve scalability of Collections by linking from Objects to Collections (instead of linking from Collections to their member Objects).

This PR removes the existing functionality to link from Collections to Objects.  Objects are typically members of one or a small number of Collections, but some Collections can have 10,000 or more member Objects.  While there is ongoing work to address the issue, Fedora performance for Collections that link to that many Objects with pcdm:hasMember is poor, so this PR dramatically improves performance by avoiding that bottleneck.


app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -115,6 +115,7 @@ def form_class
 
     def build_form
       @form = form_class.new(curation_concern, current_ability)
+      @collections = Collection.all
```
jeremyf: The return value of this method has changed, which may or may not be an issue. I'm also concerned about adding another instance variable for the view to use.

Suggestion:

1. Add form_class#collections_for_select method that can be referenced in the view

app/views/curation_concerns/base/_attributes.html.erb
```diff
@@ -9,5 +9,17 @@
     <%= presenter.attribute_to_html(:embargo_release_date, render_as: :date) %>
     <%= presenter.attribute_to_html(:lease_expiration_date, render_as: :date) %>
     <%= presenter.attribute_to_html(:rights, render_as: :rights) %>
+    <% if defined?(presenter.member_of_collection_ids) && presenter.member_of_collection_ids.size > 0 %>
```
jeremyf: You are querying `presenter.member_of_collection_ids` but down below are using `presenter.member_of_collection_presenters`. I believe the guarding if statement should reflect the method used below.

Option 1:

```ruby
if presenter.try(:member_of_collection_presenters).present?
```

Option 2:

```ruby
defined?(presenter.member_of_collection_presenters?) && presenter.member_of_collection_presenters?
```

Other options?

app/views/curation_concerns/base/_form_member_of_collections.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if @form.respond_to? :member_of_collection_ids %>
+  <div class="row" with-headroom">
+    <legend>Member of Collections</legend>
+    <%= f.input :member_of_collection_ids, collection: @collections, label_method: :first_title, input_html: { multiple: true } %>
```
jeremyf: Would prefer `@form.collections_for_select` instead.

app/actors/curation_concerns/actors/add_as_member_of_collections_actor.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Actors
+    class AddAsMemberOfCollectionsActor < AbstractActor
+      def create(attributes)
+        collection_ids = attributes.delete(:member_of_collection_ids)
+        next_actor.create(attributes) && add_to_collections(collection_ids)
+      end
+
+      def update(attributes)
+        collection_ids = attributes.delete(:member_of_collection_ids)
+        add_to_collections(collection_ids) && next_actor.update(attributes)
```
jeremyf: Is there a reason for `create` that `next_actor.create` happens before `add_to_collections` but for `update` the `next_actor.update` happens after the `add_to_collections`?

escowles:
Screenshots demonstrating new functionality:

Selecting one or more collections on the Work create or edit form:
object-new-edit

Listing collections a Work is a member of on the Work show page:
object-show

Collection membership is a facet in search results:
search-facet

app/actors/curation_concerns/actors/add_as_member_of_collections_actor.rb
escowles: I was following the pattern of the existing AddToWorkActor and AddToCollectionActor — I think the rationale for switching them might be to make sure that the Work is created before trying to add it to the Collections.  But that doesn't apply here, so I'll make them both add to Collections and then call the next Actor on the stack.

escowles:
@jeremyf thanks for your comments — I've pushed another commit to address them.

pgwillia:
This is still preliminary for us but in order to get this to work for our sufia 7.1.0 based app, in our project's Gemfile we had to include 
```
gem 'hydra-pcdm', github:'projecthydra/hydra-pcdm', branch: :master
gem 'hydra-works', github:'projecthydra/hydra-works', branch: :master
```
and make one more modification to curation_concerns
```
diff --git a/app/models/concerns/curation_concerns/collection.rb b/app/models/concerns/curation_concerns/
index 2c948c9..5b4ab12 100644
--- a/app/models/concerns/curation_concerns/collection.rb
+++ b/app/models/concerns/curation_concerns/collection.rb
@@ -9,6 +9,11 @@ module CurationConcerns
     def add_members(new_member_ids)
       return if new_member_ids.nil? || new_member_ids.empty?
       members << ActiveFedora::Base.find(new_member_ids)
+      members.each do |m|
+        m.member_of_collections << self 
+        m.member_of_collection_ids << id
+        m.save!
+      end
     end
   end
 end
```

escowles:
@pgwillia Thanks for bringing that up.  I think it's better to leave the members/add_members alone (so Collections can still have child Collections if anyone wants to do that), but I've added some helper methods like what you outlined to add a batch of objects to a collection.  See https://github.com/projecthydra/curation_concerns/pull/901/commits/b6fa6085d7601110e57d1e25b6b9fd7f8b681917

tpendragon:
We need an automated data migration method to do this.

pgwillia:
@escowles The problem that I was trying to solve is that Sufia doesn't use an Actor like your Plum does, so needed a place to populate member_of_collections. Where the collection added members seemed logical to me.  Shouldn't the inverted collection membership apply to all members regardless of class?

I'm not sure what happened to the changes you referenced in b6fa608.  I'd like to take a look at them.

pgwillia:
Sufia calls 

```
# app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
80     def create
81       @collection.apply_depositor_metadata(current_user.user_key)
82       add_members_to_collection unless batch.empty?
...
216       def add_members_to_collection(collection = nil)
217         collection ||= @collection
218         collection.add_members batch
219       end
```

pgwillia:
Did @tpendragon last commit to this branch step on some of @escowles work?

tpendragon:
Er, maybe? @escowles can you check? I thought I just got rid of the AF lock.

tpendragon:
@pgwillia Looks like @escowles pushed up the missing code.

escowles:
Are there any outstanding issues with this PR? @pgwillia @jeremyf @projecthydra/sufia-code-reviewers ? If not, I can rebase it to address the merge conflicts.

mjgiarlo:
@escowles that sounds like a no to me. Rebase away!

jeremyf:
@escowles We have a broken build due to dependencies on RDF.

jcoyne:
@escowles this last change (requiring RDF 2.1) is going to be a breaking change for folks who want the ordering behavior present in RDF 1.

escowles:
@jcoyne I know — but I thought it was unavoidable because we're requiring a version of AF that requires RDF 2.x.

But I just checked and noticed that `curation_concerns.gemspec` is only requiring AF >= 10.3.0.rc1, whereas the test app `Gemfile.extra` is requiring 11.0.0.rc7.  I'll see if I can relax the `Gemfile.extra` to match the gemspec and get it working with RDF 1.x.

spec/test_app_templates/Gemfile.extra
```diff
@@ -1,7 +1,7 @@
 # placeholder to use for pinning against specific gem commit references
 gem 'hydra-pcdm', github:'projecthydra/hydra-pcdm', branch: 'master'
 gem 'hydra-works', github:'projecthydra/hydra-works', branch: 'master'
-gem 'active-fedora', '11.0.0.rc7'
+gem 'active-fedora', '>= 10.3.0.rc1'
```
jcoyne: Why do we need this in `Gemfile.extra`? Isn't this dependency already expressed in `curation_concerns.gemspec`?

spec/test_app_templates/Gemfile.extra
```diff
@@ -1,4 +1,7 @@
 # placeholder to use for pinning against specific gem commit references
+gem 'hydra-pcdm', github:'projecthydra/hydra-pcdm', branch: 'master'
+gem 'hydra-works', github:'projecthydra/hydra-works', branch: 'master'
+gem 'active-fedora', '>= 10.3.0.rc1'
```
jcoyne: Why do we need this line in `Gemfile.extra`?  This is already expressed in `curation_concerns.gemspec`

tpendragon:
@escowles Were those two tests really unused? They seem to be around the collection show page "add member" functionality.

escowles:
@tpendragon I removed the tests because this PR removes the underlying functionality, since collection membership is now managed through the work edit form instead of through the collection edit form.

I'm not opposed to reimplementing that functionality if people think it's useful.  But this PR is already big enough I thought it was better to not add new stuff to it.

tpendragon:
@escowles I think we should maintain the current UI's functionality - it feels unexpected that it would change.

app/indexers/curation_concerns/collection_indexer.rb
```diff
@@ -10,6 +10,11 @@ def generate_solr_document
         # Index the size of the collection in bytes
         solr_doc[Solrizer.solr_name(:bytes, STORED_LONG)] = object.bytes
         solr_doc['thumbnail_path_ss'] = thumbnail_path
+
+        object.in_collections.each do |col|
+          solr_doc['member_of_collection_ids_ssim'] = col.id
```
escowles: No, this should be using `<<` instead of `=` to allow the object to be in multiple collections — I'll update that.

app/views/curation_concerns/base/_form_member_of_collections.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if @form.respond_to? :member_of_collection_ids %>
+  <div class="row" with-headroom">
```
mjgiarlo: @escowles You've either got an extraneous or a missing `"` here.

curation_concerns.gemspec
```diff
@@ -45,6 +45,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'dry-struct', '~> 0.1'
   spec.add_dependency 'flot-rails', '~> 0.0.7'
   spec.add_dependency 'highcharts-rails'
+  spec.add_dependency 'hydra-pcdm', '~> 0.9.0'
```
mjgiarlo: Instead of pinning hydra-pcdm in CC, should we update the hydra-works gemspec and let CC's dependence on HW pick up the right version of HP?

spec/test_app_templates/Gemfile.extra
```diff
@@ -1,4 +1,6 @@
 # placeholder to use for pinning against specific gem commit references
+gem 'hydra-pcdm', github:'projecthydra/hydra-pcdm', branch: 'master'
```
mjgiarlo: Are these still needed? I'd like to see these taken out before we merge this PR -- and then we can cut the 2.0 release soon thereafter.

mjgiarlo:
@escowles Would you like to squash the commits, or do you think they're good logical chunks that should remain separate? I could go either way.

app/services/curation_concerns/actors/actor_factory.rb
```diff
@@ -9,6 +9,7 @@ def self.build(curation_concern, current_user)
 
       def self.stack_actors(curation_concern)
         [AddToCollectionActor,
+         AddAsMemberOfCollectionsActor,
```
jcoyne: Why are we making this a new actor and not replacing `AddToCollectionActor`?

escowles: My intention was to leave the existing membership functionality for Collections to continue to have member Collections.  But I could just replace `AddToCollectionActor` if that makes more sense.

curation_concerns.gemspec
escowles: I've uddated https://github.com/projecthydra/hydra-works/pull/303 to make HW require hydra-pcdm `>= 0.9.0`.  I can update this PR to use that once it's released.

mjgiarlo: Merged, @escowles. Thanks! HW now needs a release and this PR can then point at the new release.

escowles:
@mjgiarlo I've rebased and squashed this PR, now depending on hydra-works 0.15.
@jcoyne Does [my answer above](https://github.com/projecthydra/curation_concerns/pull/901#discussion_r89156740) address your concern, or would you like me to replace AddToCollectionActor?

app/services/curation_concerns/actors/actor_factory.rb
jcoyne: I don't think Collections use the ActorStack, so I think we can get rid of it, right?

escowles:
@jcoyne I've removed the AddToCollectionActor to avoid that duplication.

app/models/concerns/curation_concerns/collection.rb
```diff
@@ -10,5 +10,17 @@ def add_members(new_member_ids)
       return if new_member_ids.nil? || new_member_ids.empty?
       members << ActiveFedora::Base.find(new_member_ids)
     end
+
+    def add_member_objects(new_member_ids)
+      Array(new_member_ids).each do |member_id|
+        member = ActiveFedora::Base.find(member_id, cast: true)
```
jcoyne: The `cast: true` doesn't do anything (anymore). please remove that.

app/models/concerns/curation_concerns/collection.rb
```diff
@@ -10,5 +10,17 @@ def add_members(new_member_ids)
       return if new_member_ids.nil? || new_member_ids.empty?
       members << ActiveFedora::Base.find(new_member_ids)
     end
+
+    def add_member_objects(new_member_ids)
```
jcoyne: Add some documentation here about how this adds the association on the member side. If you wouldn't mind, please also add some documentation to `add_members` too.

app/presenters/curation_concerns/work_show_presenter.rb
```diff
@@ -84,8 +82,12 @@ def composite_presenter_class
 
     # @return [Array<CollectionPresenter>] presenters for the collections that this work is a member of
     def collection_presenters
-      PresenterFactory.build_presenters(in_collection_ids,
-                                        collection_presenter_class,
+      PresenterFactory.build_presenters(in_collection_ids, collection_presenter_class,
+                                        *presenter_factory_arguments)
+    end
+
+    def member_of_collection_presenters
```
jcoyne: Add method documentation here.

app/presenters/curation_concerns/work_show_presenter.rb
```diff
@@ -84,8 +82,12 @@ def composite_presenter_class
 
     # @return [Array<CollectionPresenter>] presenters for the collections that this work is a member of
     def collection_presenters
```
jcoyne: Is this method still needed?

app/views/catalog/_action_menu_partials/_collection.html.erb
```diff
@@ -7,17 +7,13 @@
...
-                          <i class="glyphicon glyphicon-trash"></i> Delete <%= document.human_readable_type %>
-          <% end %>
+        <%= link_to document, class: 'itemicon itemtrash', title: 'Delete Collection', method: :delete, data: {
+                        confirm: "Deleting a collection from #{application_name} is permanent. Click OK to delete this collection from #{application_name}, or Cancel to cancel this operation" } do %>
+                        <i class="glyphicon glyphicon-trash"></i> Delete <%= document.human_readable_type %>
         <% end %>
       </li>
     <% end %>
-    <% if can? :collect, document %>
+    <% if can?(:collect, document) && document.collection? %>
```
jcoyne: Why is this check required. Isn't this used on a show page for a collection (which ensures that document is a Collection)?

app/views/collections/edit.html.erb
```diff
@@ -1,7 +1,6 @@
 <% provide :page_title, curation_concern_page_title(@collection) %>
 
-<h1>Edit <%= @collection %> <span class="human_readable_type">(<%= @collection.human_readable_type %>)</span></h1>
-
+<h1>Edit <%= @collection.first_title %> <span class="human_readable_type">(<%= @collection.human_readable_type %>)</span></h1>
```
jcoyne: use `@collection.to_s` for the title rather than `first_title`.

app/views/curation_concerns/base/_form_member_of_collections.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if @form.respond_to? :member_of_collection_ids %>
```
jcoyne: Why do we need to check this?  Shouldn't it always be present?

app/views/curation_concerns/base/_member_of_collections.html.erb
```diff
@@ -0,0 +1,12 @@
+    <% if defined?(presenter.member_of_collection_presenters) && presenter.member_of_collection_presenters.size > 0 %>
```
jcoyne: Why do we need to check for `defined?(presenter.member_of_collection_presenters)`?

app/views/curation_concerns/base/_member_of_collections.html.erb
```diff
@@ -0,0 +1,12 @@
...
+      <tr>
+        <td>Member of Collections</td>
+        <td>
+          <ul class="member_of_collections">
+            <% presenter.member_of_collection_presenters.each do |p| %>
+              <li><%= link_to p.title.first, main_app.collection_path(p.id), class: 'collection-link' %></li>
```
jcoyne: Can we encapsulate `p.title.first` in a method (`#to_s` perhaps)?

app/views/catalog/_action_menu_partials/_collection.html.erb
jcoyne: Is this template used at all?

escowles:
@jcoyne I've added some docs and removed some unnecessary methods and checks leftover from when I was trying to support both the member and member_of membership modes.

app/views/curation_concerns/base/_member_of_collections.html.erb
```diff
@@ -0,0 +1,12 @@
+<% if defined?(presenter.member_of_collection_presenters) && presenter.member_of_collection_presenters.size > 0 %>
```
jcoyne: Yes. Let's get rid of this check here and put it in the `_attributes` partial.

jcoyne:
I'm dismissing @mjgiarlo's review because all those changes have been made.

PR #911, opened 2016-08-17
Use params keyword with ActionController::TestCase

### LINKED ISSUE #905: ActionController::TestCase needs keyword args
```
DEPRECATION WARNING: ActionController::TestCase HTTP request methods will accept only
keyword arguments in future Rails versions.

Examples:

get :show, params: { id: 1 }, session: { user_id: 1 }
process :update, method: :post, params: { id: 1 }
```

Occurs hundreds of times in test suite.

### LINKED ISSUE #907: :nothing option deprecated
```
DEPRECATION WARNING: `:nothing` option is deprecated and will be removed in Rails 5.1. 
Use `head` method to respond with empty response body. 
(called from block (4 levels) in <top (required)> at
 /Users/atz/repos/curation_concerns/spec/controllers/downloads_controller_spec.rb:10)
```

### LINKED ISSUE #910: xhr deprecated
```
DEPRECATION WARNING: 
xhr and xml_http_request methods are deprecated in favor of `get :index, xhr: true` and `post :create, xhr: true`
(called from block (5 levels) in <top (required)> at
 .../spec/controllers/curation_concerns/file_sets_controller_spec.rb:98)
```

atz:
Fixes #910
Fixes #907 
Largely fixes #905.  Remaining TestCase warnings are from a helper class.

Saves hundreds of lines of deprecation warnings per CI run.

Converts `xhr` calls to their corresponding HTTP method with `xhr: true`.  (#910)

### DIFF
--- a/spec/support/backport_test.rb
+++ b/spec/support/backport_test.rb
@@ -1,0 +1,14 @@
+# Backport the Rails 5 controller test methods to Rails 4
+module BackportTest
+  [:delete, :get, :post, :put, :patch].each do |http_action|
+    define_method(http_action) do |*args|
+      (action, rest) = *args
+      rest ||= {}
+      if rest[:xhr]
+        @request.env['HTTP_X_REQUESTED_WITH'] = 'XMLHttpRequest'
+        @request.env['HTTP_ACCEPT'] ||= [Mime::JS, Mime::HTML, Mime::XML, 'text/xml', Mime::ALL].join(', ')
+      end
+      super(action, rest[:params])
+    end
+  end
+end
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -35,18 +35,16 @@ describe CatalogController do
 
     context 'Searching all works' do
       it 'returns all the works' do
-        get 'index', 'f' => { 'generic_type_sim' => 'Work' }
+        get 'index', params: { 'f' => { 'generic_type_sim' => 'Work' } }
         expect(response).to be_successful
         expect(assigns(:document_list).count).to eq 2
-        [work1.id, work2.id].each do |work_id|
-          expect(assigns(:document_list).map(&:id)).to include(work_id)
-        end
+        expect(assigns(:document_list).map(&:id)).to include(work1.id, work2.id)
       end
     end
 
     context 'Searching all collections' do
       it 'returns all the works' do
-        get 'index', 'f' => { 'generic_type_sim' => 'Collection' }
+        get 'index', params: { 'f' => { 'generic_type_sim' => 'Collection' } }
         expect(response).to be_successful
         expect(assigns(:document_list).map(&:id)).to eq [collection.id]
       end
@@ -54,7 +52,7 @@ describe CatalogController do
 
     context 'searching just my works' do
       it 'returns just my works' do
-        get 'index', works: 'mine', 'f' => { 'generic_type_sim' => 'Work' }
+        get 'index', params: { works: 'mine', 'f' => { 'generic_type_sim' => 'Work' } }
         expect(response).to be_successful
         expect(assigns(:document_list).map(&:id)).to eq [work1.id]
       end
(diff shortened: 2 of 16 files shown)


jcoyne:
I suspect this will break for Rails 4, but you could copy this from Blacklight, which will fix it: https://github.com/projectblacklight/blacklight/commit/59436b48828c96a013a207fab079d82d4bc2c793#diff-93830fa29d616f7c87903d08b5b1b29aR70

atz:
I thought the `params` part is old.  You think the `xhr: true` will break it?

atz:
Yep, you're right about the break in Rails 4.  Thx for the tip, will fold that in.

atz:
You know, for how fugly the specs are and the _dozens_ of rubocop exceptions CC has adopted, breaking the build on an extra line break seems pretty damn rich.

jcoyne:
@atz create an executable (chmod 755) file in `.git/hooks/pre-commit`
```
#!/bin/sh

exec bundle exec rubocop
```

Then you'll never have that problem again.

atz:
It was from your lines I copied over from Sufia, for the record.

atz:
@jcoyne Any feedback on why the backport workaround isn't satisfying the routes under rails 4?
https://travis-ci.org/projecthydra/curation_concerns/jobs/152870163

Did I miss pulling something in?  Still works fine on rails 5.  Ah... looks like it is from `patch` requests that didn't get intercepted.

atz:
Ok, it's down to 4x routing error:
```
ActionController::UrlGenerationError:
       No route matches {:action=>"show", :controller=>"curation_concerns/file_sets"}
```
And one `ActionController::UnknownFormat`.

https://travis-ci.org/projecthydra/curation_concerns/jobs/153150727

atz:
```ruby
respond_to do |format|
  format.html do
    if request.xhr?
      render 'jq_upload', formats: 'json', content_type: 'text/html'
    else
      redirect_to [main_app, curation_concern.parent]
    end
  end
  format.json do
    render 'jq_upload', status: :created, location: polymorphic_path([main_app, curation_concern])
  end
end
```

XHR... requests html, give it json with html mime-type?  Is that the logic I'm seeing there?

jcoyne:
@atz yeah, it doesn't make a lot of sense, but I believe it's due to one of the javascript libraries/browsers not setting the Accept type correctly on XHR requests.

atz:
Probably deserves a comment to that effect, at least.  Would be nice to know to be able to rip it out when that lib is updated/replaced/removed.

atz:
Got it down to just that one failure.

jcoyne:
@atz one thing the old XHR method did, that our method is not doing is set the HTTP_ACCEPT header as seen here:
https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_controller/test_case.rb#L549

Perhaps you can modify that support file to add that line?

jcoyne:
Or even change it to:
```ruby
if rest[:xhr]
  xml_http_request(:post, action, rest[:params])
else
 super(action, rest[:params])
end
```

atz:
Yeah, the extra ENV var does it.  Seeing as HTTP_ACCEPT is so critical to server response, I'm not surprised that leaving it out would break something.  I collapsed the backports file down to 12 lines with the changes.  We probably want to use that version in blacklight, too, since it adds that ENV and `patch` support.

Thanks for the insights.

spec/support/backport_test.rb
```diff
@@ -0,0 +1,12 @@
+# Backport the Rails 5 controller test methods to Rails 4
+module BackportTest
+  [:delete, :get, :post, :put, :patch].each do |http_action|
+    define_method(http_action) do |*args|
+      (action, rest) = *args
+      rest ||= {}
+      @request.env['HTTP_X_REQUESTED_WITH'] = 'XMLHttpRequest' if rest[:xhr]
+      @request.env['HTTP_ACCEPT'] ||= [Mime::JS, Mime::HTML, Mime::XML, 'text/xml', Mime::ALL].join(', ')
```
jcoyne: I think this should only be set `if rest[:xhr]`

atz: Probably doesn't matter since all likely targets are covered, literally including Mime::ALL, but sure.

PR #918, opened 2016-08-18
Add sortable_placeholder from plum, fixes #915

### LINKED ISSUE #915: file manager drag/drop dead spot
### Descriptive summary

cc1.4

If I have 4 children of a work (in file manager, 1 full row and 1 item on the 2nd row)
I cannot drag the last object into position 3. I also cannot drag the 3rd object into position 4.

hackmastera:
Creates a space when you drag an object in the file-manager so you can put an object in the last position. also keeps footer from jumping up when there was just 1 object in the final row.

### DIFF
--- a/app/assets/javascripts/curation_concerns/file_manager.es6
+++ b/app/assets/javascripts/curation_concerns/file_manager.es6
@@ -7,6 +7,7 @@ export default class FileManager {
     this.sorting()
     this.save_affix()
     this.member_tracking()
+    this.sortable_placeholder()
   }
 
   initialize_save_manager() {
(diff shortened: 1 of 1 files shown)


mjgiarlo:
@hackmastera could use a test that fails without `sortable_placeholder` but if that gets too annoying, I'd be happy to merge without it. I don't see any specific tests in Plum for this.

app/assets/javascripts/curation_concerns/file_manager.es6
```diff
@@ -38,4 +39,12 @@ export default class FileManager {
       $(element).data("file_manager_member", manager_member)
     })
   }
+
+  sortable_placeholder() {
+    $( "#sortable" ).on( "sortstart", function( event, ui ) {
```
jcoyne: Please add a comment here about what is going on and why.

hackmastera:
inserted explanatory comment at @jcoyne's request. Perhaps @tpendragon would like to fact-check it.

tpendragon:
Fact-checked, :+1: Thanks for getting this in @hackmastera

tpendragon:
@jcoyne @mjgiarlo Look good to you?

PR #919, opened 2016-08-18
Refactor AdminSetService to accept the controller context

cbeer:


app/services/curation_concerns/admin_set_service.rb
```diff
@@ -1,7 +1,11 @@
 module CurationConcerns
   class AdminSetService
-    def initialize(user)
+    attr_reader :user, :context
+    delegate :blacklight_config, :repository, to: :context
+
+    def initialize(user, context)
```
atz: Probably better as `def initialize(context, user)`, with the more volatile argument second.  

I think this looks better:
```ruby
AdminSetService.new(self, current_user)
```
than
```ruby
AdminSetService.new(current_user, self)
```

All search services will need a context, not all will need a user.

atz: @cbeer I'm proceeding w/ `CollectionService` based on this branch and the proposed switched param order.

cbeer: 👌 , please go ahead and fix the order on this branch.

jcoyne: I wonder why we need two arguments at all. Presumably the `#current_user` is available on the `context`.

cbeer: If it is, great (and then we can delegate `current_ability` to the context); I had assumed passing in the user was a feature.

PR #923, opened 2016-08-19
Adding characterization terms to the solr indexes and to the FileSet presenter

cam156:
This is to fix https://github.com/projecthydra/sufia/issues/2431 in Sufia.

I put a PR into sufia (https://github.com/projecthydra/sufia/pull/2487) and was asked to move it to curation concerns.

This adds the characterization terms to the file set indexer, the solr document and more characterization terms to the fileset presenter.

### DIFF
--- a/app/indexers/curation_concerns/file_set_indexer.rb
+++ b/app/indexers/curation_concerns/file_set_indexer.rb
@@ -22,6 +22,11 @@ module CurationConcerns
         # between files on disk (in fcrepo.binary-store-path) and objects
         # in the repository.
         solr_doc[Solrizer.solr_name('digest', :symbol)] = digest_from_content
+        solr_doc[Solrizer.solr_name('page_count')] = object.page_count
+        solr_doc[Solrizer.solr_name('file_title')] = object.file_title
+        solr_doc[Solrizer.solr_name('duration')] = object.duration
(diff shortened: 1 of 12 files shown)


app/indexers/curation_concerns/file_set_indexer.rb
```diff
@@ -22,6 +22,12 @@ def generate_solr_document
         # between files on disk (in fcrepo.binary-store-path) and objects
         # in the repository.
         solr_doc[Solrizer.solr_name('digest', :symbol)] = digest_from_content
+        solr_doc[Solrizer.solr_name('page_count')] = object.page_count
+        solr_doc[Solrizer.solr_name('file_title')] = object.file_title
+        solr_doc[Solrizer.solr_name('duration')] = object.duration
+        solr_doc[Solrizer.solr_name('sample_rate')] = object.sample_rate
+        # solr_doc[Solrizer.solr_name('last_modified')] = object.last_modified
```
jcoyne: Cruft?

cam156: I was thinking last_modified should be something we could index, but it seems to not be in Hydra::PCDM::File.  I'm not 100% sure if we should just ignore, add a ticket into Hydra::Works, or if there is another field in Hydra::Works that is last_modified

PR #925, opened 2016-08-19
Add 'flip' for dynamically toggling features on and off

jcoyne:


jcoyne:
Looks like I need a test for the `Admin::StrategiesController`.

jcoyne:
Okay, test added, good to merge.

cbeer:
Should we wait for your PR to flip to be accepted, at least?

jpstroop:
Can you explain? Is the idea that features could be en/disabled anywhere further up the stack and that the core functionality to do so lives in CC, or are you imagining other features in CC for which this could be possible (or both)?

jcoyne:
> Is the idea that features could be en/disabled anywhere further up the stack and that the core functionality to do so lives in CC

Yes, for now this is the case.  In the long term, for sustainability of the project, I think we need to merge sufia and cc, and this is one mechanism for enabling that.

jpstroop:
Thanks. That makes sense.

> [...] we need to merge sufia and cc, and this is one mechanism for enabling that.

This isn't the right venue obviously, but one concern I have about lumping everything together is that features some stakeholders don't need could wind up introducing bugs and slowing down releases for others who don't. I do hear your point about sustainability though, so we probably should get to work enumerating the features and where they do/don't overlap.

Also need to consider other projects like GeoConcerns and how they'd fit in.

awead:
I like this, but my only reservation is the Flip gem. It looks like @jcoyne has put in some PRs but is working off of a fork. Do we have any inclination that the gem's maintainer would be receptive to our PRs? I'd rather have using a released version then have to use a fork.

tpendragon:
If this is a requested feature (is this PR actually doing anything? There doesn't seem like much logic for what a feature "is" and how it should be locked out) then we should talk about it on hydra-tech so we know the use case. At the moment I'm inclined to say "if the vast majority of people don't want single use links, they shouldn't be in curation concerns."

If, however, this is an example application of flip and it's being integrated as a stepping stone to merging Sufia with Curation Concerns so we have only one repository, I think the complexity should wait until the architecture group has come to a recommendation about that.

jcoyne:
@tpendragon the feature in this PR is "the ability to dynamically configure features in the application".  One of those features is a UI for admin sets.  That is not the only feature that people have asked to enable and disable, but it is on that is immediately comes to mind.  There are many institutions who want admin sets (See WUSTL, UCSB, Stanford, Hull. etc.). I also know of a number of institutions who don't want them (PSU comes to mind) and several that are using User Collections as Admin Sets (and are feeling the pain of Fedora's slowness there). So I think we can say there is significant interest.  I don't think we need to have everyone want a feature in order to include it.

escowles:
I think the target audience for CC is people building their own applications tailored to their local needs.  The target audience for flip (and feature-toggling in general) is building an application that users can tailor to their own needs.

For AdminSets, I think having a significant number of people interested is enough to include it in CC.  If a small number of people don't want it, they should be able to easily override to hide it in their app and/or suppress the behavior.  If there is less agreement, then having it in a separate gem that can be included or not seems like the way to go.

mjgiarlo:
:speech_balloon: @jpstroop 

> ... features some stakeholders don't need could wind up introducing bugs and slowing down releases for others who don't.

No contest re: bugs. This is true in all our codebases, eh?

Re: slowing down releases, well, that depends. You could say that the Sufia 7.0.0 work was a good example of slowing down releases, but folks using 6.x were able to continue pushing releases thanks to branches. Another variable here is how much capacity we have to expedite the work needed to make a release, and we'd have greater capacity if the Sufia and CC developers were working on the same codebase.

PR #926, opened 2016-08-19
Allow thumbnail to be selected from the file manager

tpendragon:
There's now a thumbnail radio button which spans all the individual boxes in the file manager. It looks like this:

screen shot 2016-08-19 at 2 10 37 pm

### DIFF
--- a/app/assets/javascripts/curation_concerns/file_manager.es6
+++ b/app/assets/javascripts/curation_concerns/file_manager.es6
@@ -1,6 +1,6 @@
 import SaveManager from 'curation_concerns/file_manager/save_manager'
 import SortManager from 'curation_concerns/file_manager/sorting'
-import {FileManagerMember} from 'curation_concerns/file_manager/member'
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/curation_concerns/file_manager.es6
```diff
@@ -40,6 +41,18 @@ export default class FileManager {
     })
   }
 
+  resource_form() {
```
jcoyne: Please add a comment about what's happening here.

app/services/curation_concerns/indexes_thumbnails.rb
```diff
@@ -13,6 +13,7 @@ module IndexesThumbnails
     def generate_solr_document
       super.tap do |solr_doc|
         index_thumbnails(solr_doc)
+        solr_doc[Solrizer.solr_name('thumbnail_id', :symbol)] = object.thumbnail_id
```
jcoyne: It seems like this would fit better in `index_thumbnails`

awead:
LGTM. 👍

PR #928, opened 2016-08-22
Add publication_state predicate

jcoyne:
This would be the model for projecthydra/sufia#2452

### DIFF
--- a/spec/models/generic_work_spec.rb
+++ b/spec/models/generic_work_spec.rb
@@ -14,6 +14,28 @@ describe GenericWork do
     it { is_expected.to eq 'curation_concerns_generic_work' }
   end
 
+  describe '#state' do
+    let(:work) { described_class.new(state: inactive) }
+    let(:inactive) { ::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive') }
+    subject { work.state.rdf_subject }
+    it { is_expected.to eq inactive }
+  end
+
+  describe '#suppressed?' do
+    let(:work) { described_class.new(state: state) }
+    subject { work.suppressed? }
+
+    context "when the state is 'pending'" do
+      let(:state) { 'pending' }
+      it { is_expected.to be true }
+    end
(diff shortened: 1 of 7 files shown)


spec/models/generic_work_spec.rb
```diff
@@ -14,6 +14,16 @@
     it { is_expected.to eq 'curation_concerns_generic_work' }
   end
 
+  describe '#object_state' do
+    let(:work) { described_class.new }
+    let(:inactive) { ::RDF::URI('http://fedora.info/definitions/1/0/access/ObjState#inactive') }
```
cbeer: Not a blocker, but should we get these URIs pushed into `rdf-vocab` or similar?

jcoyne: I doubt that this is going to get published in `rdf-vocab` as it's not a widely used vocabulary.

mjgiarlo: Not *necessarily* true. We've already got https://github.com/ruby-rdf/rdf-vocab/blob/develop/lib/rdf/vocab/fcrepo4.rb in there, and there are a handful of Fedora/Hydra folks shepherding that gem along.

app/models/vocab/fedora_state.rb
```diff
@@ -0,0 +1,6 @@
+require 'rdf'
+module Vocab
+  class FedoraState < RDF::StrictVocabulary('http://fedora.info/definitions/1/0/access/')
```
mjgiarlo: Minor points, but:

* Should we align the name of this class with the name of the ontology? It's the Fedora Resource Status ontology.
* I believe the URI here may be wrong. you're missing `ObjState#` on the end.

jcoyne: I copied this class in from https://github.com/pulibrary/plum/blob/master/app/models/vocab/f3_access.rb#L2, so @escowles be aware that you may have the same issue.

PR #930, opened 2016-08-24
Changing version revision to use the correct signature for CharacterizeJob

cam156:
Fixes https://github.com/projecthydra/sufia/issues/2404

The signature for characterization job changed a while back and this update was missed becuase the test is completely mocked.

I changed the mocking to the correct signature, but I am not certain that the test is really testing all that we want it to test.  Do people have ideas on how to better test the integration without writing an integration test?

### DIFF
--- a/app/actors/curation_concerns/actors/file_actor.rb
+++ b/app/actors/curation_concerns/actors/file_actor.rb
@@ -34,9 +34,8 @@ module CurationConcerns
 
         CurationConcerns::VersioningService.create(repository_file, user)
 
-        # Retrieve a copy of the original file from the repository
-        working_file = WorkingDirectory.copy_repository_resource_to_working_directory(repository_file, file_set.id)
-        CharacterizeJob.perform_later(file_set, working_file)
+        # Characterize the original file from the repository
+        CharacterizeJob.perform_later(file_set, repository_file.id)
         true
       end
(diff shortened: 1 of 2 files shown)


awead:
I suspect it didn't blow up because "working_path" in the old signature and "file_set.id" in the new sig. were just strings, so the test didn't see a difference. I'm not sure if there really is a way around this other than to have an integration test.

But, something doesn't look right about that second parameter. You're passing `file_set.id` except that's supposed to be an "identifier for a Hydra::PCDM::File" according to the documentation. In the other classes that are invoked by CharacterizeJob, it's also expecting an identifier for a Hydra::PCDM::File:

https://github.com/projecthydra/curation_concerns/blob/master/app/services/curation_concerns/working_directory.rb#L7

https://github.com/projecthydra/curation_concerns/blob/master/app/jobs/create_derivatives_job.rb#L6

cam156:
@awead good catch!

awead:
@cam156 LGTM. I still don't have a good answer for the test coverage question, though. I'm good with merging if this solves the current problem, even if it doesn't protect us from it happening again in the future.

cam156:
@awead it does fix the revision reversion issue in Sufia (I tested that locally), so merging it seems like a good idea.

PR #932, opened 2016-08-24
Add RightsStatements service

jcoyne:
This is preliminary to adding `edm:rights` which will hold the Rights Statement assertion on the form.


mjgiarlo:
Thoughts, @tpendragon @escowles @jpstroop ?

escowles:
This is great!  A nice extension of this would be to also include the scope note and descriptive text of the rights statements, and some methods in the service to work with those (see https://github.com/pulibrary/plum/blob/master/config/authorities/rights.yml#L25-L40 and https://github.com/pulibrary/plum/blob/master/app/services/rights_statement_service.rb#L4-L7).  That would enable applications to include the user-facing definitions of the rights statements to show to users.

PR #937, opened 2016-08-26
Supply the main_app route set

jcoyne:
This enables sufia to call this partial when isolating the engine
namespace


app/views/collections/_button_create_collection.html.erb
```diff
@@ -1,2 +1,2 @@
-<%= button_to label, new_collection_path, :method=>:get, :class=>"btn btn-primary collection-add submits-batches", 'data-behavior'=>'hydra-collections', :id=>'hydra-collection-add' %>
+<%= button_to label, main_app.new_collection_path, :method=>:get, :class=>"btn btn-primary collection-add submits-batches", 'data-behavior'=>'hydra-collections', :id=>'hydra-collection-add' %>
```
mjgiarlo: :hash: :rocket: :boom: :question: ;)

PR #938, opened 2016-08-26
Unset file manager form callbacks after they run so they don't build up

hackmastera:
pull request courtesy of hydra code club


jcoyne:
:+1:

PR #940, opened 2016-08-26
Override #local_prefixes instead of private #_prefixes method

### LINKED ISSUE #939: Use local_prefixes instead of internal _prefix
### Descriptive summary

We're overriding `_prefixes` in https://github.com/projecthydra/curation_concerns/blob/master/app/controllers/concerns/curation_concerns/curation_concern_controller.rb#L138

But as of Rails 4.2, there's `local_prefixes which may be a better option.

cbeer:
Fixes #939


tpendragon:
Rubocop.

mjgiarlo:
@tpendragon :speech_balloon: 
> Rubocop.

:cop: :hand: :rage1:

PR #944, opened 2016-08-29
Refactor ingest workflow to use current working file

awead:
Refs projecthydra/sufia#2553

* FileActor only downloads a working file unless the uploaded file is unavailable locally
* IngestFileJob now accepts optional parameters, such as filename, which had been causing unnecessary downloads because the tempfile had a different name than the original.
* The current working file is passed from IngestFileJob to successive jobs like CharactizeJob and CreateDerivativeJob

This avoids the problem of downstream applications forcing another download of the file because the working file may be in a different, yet local, location. In this case, once the job has the local file "in hand," it passes it to successive jobs as an optional argument. If the job receives the local file argument, it users CurationConcerns::WorkingDirectory to check if the file still exists and re-downloads it necessary. 

### DIFF
--- a/app/jobs/ingest_file_job.rb
+++ b/app/jobs/ingest_file_job.rb
@@ -2,36 +2,42 @@ class IngestFileJob < ActiveJob::Base
   queue_as CurationConcerns.config.ingest_queue_name
 
   # @param [FileSet] file_set
-  # @param [String] filename the cached file within the CurationConcerns.config.working_path
-  # @param [String,NilClass] mime_type
+  # @param [String] filepath the cached file within the CurationConcerns.config.working_path
   # @param [User] user
-  # @param [String] relation ('original_file')
-  def perform(file_set, filename, mime_type, user, relation = 'original_file')
-    local_file = File.open(filename, "rb")
+  # @option opts [String] mime_type
+  # @option opts [String] filename
+  # @option opts [String] relation, ex. :original_file
+  def perform(file_set, filepath, user, opts = {})
+    mime_type = opts.fetch(:mime_type, nil)
+    filename = opts.fetch(:filename, File.basename(filepath))
+    relation = opts.fetch(:relation, :original_file).to_sym
+    local_file = File.open(filepath, "rb")
+
     # If mime-type is known, wrap in an IO decorator
     # Otherwise allow Hydra::Works service to determine mime_type
     if mime_type
       local_file = Hydra::Derivatives::IoDecorator.new(local_file)
       local_file.mime_type = mime_type
-      local_file.original_name = File.basename(filename)
+      local_file.original_name = filename
     end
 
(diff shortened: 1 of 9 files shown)


awead:
Comments welcome. I'm looking at how this will solve the referenced issue in Sufia.

app/services/curation_concerns/working_directory.rb
```diff
@@ -1,10 +1,12 @@
 module CurationConcerns
-  module WorkingDirectory
+  class WorkingDirectory
```
jcoyne: This class is never instantiated, was this change necessary?

app/jobs/ingest_file_job.rb
```diff
@@ -2,36 +2,42 @@ class IngestFileJob < ActiveJob::Base
...
                                         local_file,
-                                        relation.to_sym,
+                                        relation,
                                         versioning: false)
 
     # Persist changes to the file_set
     file_set.save!
 
-    repository_file = file_set.send(relation.to_sym)
+    repository_file = file_set.send(relation)
```
jeremyf: Are we calling a public method? If so, prefer `#public_send` over `#send`.

app/jobs/ingest_file_job.rb
```diff
@@ -2,36 +2,42 @@ class IngestFileJob < ActiveJob::Base
...
-  # @param [String] relation ('original_file')
-  def perform(file_set, filename, mime_type, user, relation = 'original_file')
-    local_file = File.open(filename, "rb")
+  # @option opts [String] mime_type
+  # @option opts [String] filename
+  # @option opts [String] relation, ex. :original_file
+  def perform(file_set, filepath, user, opts = {})
+    mime_type = opts.fetch(:mime_type, nil)
+    filename = opts.fetch(:filename, File.basename(filepath))
+    relation = opts.fetch(:relation, :original_file).to_sym
```
jeremyf: Is `#to_sym` required? Would `#to_s` suffice? Creating symbols prior to Ruby 2.2 (or so) means creating objects that are not cleaned up.

awead: @jeremyf yeah, because ActiveJob doesn't allow symbols as parameters unfortunately.

app/services/curation_concerns/working_directory.rb
awead: In case anyone wants to extend it. It was leftover from an earlier attempt with Sufia integration when I was trying to create:
`Sufia::WorkingDirectory < CurationConcerns::WorkingDirectory` I'm ambivalent about it now.

app/jobs/ingest_file_job.rb
jcoyne: Prior to this PR it was already doing `.to_sym`, so let's not address this now.

app/jobs/ingest_file_job.rb
jcoyne: Good suggestion, but not required in this PR as this is the way it was before.

PR #945, opened 2016-08-29
Add property for rights_statement

jcoyne:
Deprecate the old `rights` property and rename it to `license` to
avoid confusion

### DIFF
--- a/app/models/concerns/curation_concerns/basic_metadata.rb
+++ b/app/models/concerns/curation_concerns/basic_metadata.rb
@@ -28,9 +28,17 @@ module CurationConcerns
       property :keyword, predicate: ::RDF::Vocab::DC11.relation do |index|
         index.as :stored_searchable, :facetable
       end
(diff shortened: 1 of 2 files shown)


mjgiarlo:
Should we discuss this on Wednesday's Hydra Tech call?

escowles:
👍  to adding `rights_statement` with the predicate `edm:rights` — that lines up with the [rights metadata recommendation](https://wiki.duraspace.org/display/hydra/Rights+Metadata+Recommendation).

I understand wanting to rename `rights` to avoid confusion with `rights_statement`, but I think that will lead to confusion with `dct:license` and might make people think they couldn't use CC licenses should go there instead of in `rights_statement`.  So I think leaving `rights` unchanged is probably better than renaming to `license`.

jcoyne:
@escowles okay, I've paired this back to just a new `rights_statement` property

PR #947, opened 2016-08-29
Add alphabetical sort to file manager

hackmastera:
Here's my approach; any comments, @tpendragon or @jcoyne ?

Thanks!

### DIFF
--- a/app/assets/javascripts/curation_concerns/file_manager/sorting.es6
+++ b/app/assets/javascripts/curation_concerns/file_manager/sorting.es6
@@ -5,6 +5,7 @@ export default class SortManager {
     this.initialize_sort()
     this.element.data("current-order", this.order)
     this.save_manager = save_manager
+    this.initialize_alpha_sort_button()
   }
 
   initialize_sort() {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/curation_concerns/file_manager/sorting.es6
```diff
@@ -82,4 +87,40 @@ export default class SortManager {
       }
     ).toArray()
   }
+
+  get alpha_sort_button() {
+    return $("*[data-action='alpha-sort-action']")
+  }
+
+  initialize_alpha_sort_button() {
+    Blacklight.onLoad(() => {
```
tpendragon: I don't think the sorting manager gets triggered until Blacklight's loaded, so this shouldn't be necessary.

app/assets/javascripts/curation_concerns/file_manager/sorting.es6
```diff
@@ -82,4 +87,40 @@ export default class SortManager {
...
+      )
+    })
+    // sort array by title of each object
+    array.sort(function(o1, o2) {
+      let a = o1.title.toLowerCase()
+      let b = o2.title.toLowerCase()
+      return a < b ? -1 : (a > b ? 1 : 0);
+    });
+    // replace contents of #sortable with elements from the array
+    this.element.empty()
```
tpendragon: Does sorting still work after you re-order this way? I'm not sure how grumpy jquery-ui-sortable gets if you delete its container contents.

hackmastera: manually confirmed that there's not a problem with this.

app/assets/javascripts/curation_concerns/file_manager/sorting.es6
hackmastera: Actually you're right and this was causing a bug that I hadn't noticed where the button didn't work unless you reload the page.

mjgiarlo:
:Tada:

PR #954, opened 2016-09-01
Revert "Refactor ingest workflow to use current working file"

jcoyne:
This also has a commit with a test that was broken when the refactor was added.


jcoyne:
@awead this reverts your work.  Are you okay with doing that until you can resolve the new test I added in this commit??

PR #961, opened 2016-09-06
Allow public users to see public collections

jcoyne:
Previously it was requiring a login to see any of the collections index view


mjgiarlo:
Failure on Travis was unrelated to work in branch -- it's a periodic failure to download Solr 6.2.0 from the mirror.

PR #962, opened 2016-09-06
Indexing for related file properties

hackmastera:
- Deduplicate thumbnail_id (remove new solr field `thumbnail_id` in favor of previously-in-use `hasRelatedImage`)
- Add representative_id to work indexer (using already-named `hasRelatedMediaFragment` field)

Pairing these commits in one PR because the fields were set up symmetrically and are functionally similar; also both are related to file manager work.

### DIFF
--- a/app/models/concerns/curation_concerns/solr_document_behavior.rb
+++ b/app/models/concerns/curation_concerns/solr_document_behavior.rb
@@ -66,7 +66,7 @@ module CurationConcerns
     end
 
     def thumbnail_id
-      first(Solrizer.solr_name('thumbnail_id', :symbol))
+      first(Solrizer.solr_name('hasRelatedImage', :symbol))
     end
 
     # Date created is indexed as a string. This allows users to enter values like: 'Circa 1840-1844'
--- a/app/presenters/curation_concerns/work_show_presenter.rb
+++ b/app/presenters/curation_concerns/work_show_presenter.rb
@@ -38,7 +38,8 @@ module CurationConcerns
     # Metadata Methods
     delegate :title, :date_created, :date_modified, :date_uploaded, :description,
              :creator, :contributor, :subject, :publisher, :language, :embargo_release_date,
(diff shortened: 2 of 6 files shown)


app/indexers/curation_concerns/work_indexer.rb
```diff
@@ -8,6 +8,7 @@ def generate_solr_document
         solr_doc[Solrizer.solr_name('member_ids', :symbol)] = object.member_ids
         Solrizer.set_field(solr_doc, 'generic_type', 'Work', :facetable)
         solr_doc[Solrizer.solr_name('suppressed', STORED_BOOL)] = object.suppressed?
+        solr_doc[Solrizer.solr_name('hasRelatedMediaFragment', :symbol)] = object.representative_id
```
jcoyne: Are you sure that this isn't already being indexed by virtue of it being an ActiveFedora relationship?

hackmastera: Didn't think of that. I know it's explicitly indexed in the file_set_indexer. Actually you may be right, though.

jcoyne: If this is the case, please remove this line.

hackmastera: okay yes both fields are coming through AF for works, but not for fileset. I'm not really sure why, or how to tell what other classes will get it through AF.

Esp. regarding hasRelatedImage, when @tpendragon added it to the indexes_thumbnails service, that propagates to a bunch of places. This is a duplication when it comes to works and I could take it out of the service and put it directly on the fileset indexer but is it also needed in the collection and admin set indexers and if so do you prefer it in the service even though we don't need it for the work?

hackmastera:
I took the thumbnail_id field out of the indexes_thumbnail service. this deduplicates on the work but means that thumbnail_id is not indexed on adminset or collection. I think this is okay because it was only added in there in the past few weeks, and I think getting it on work was the primary intention. @jcoyne @tpendragon thoughts?

PR #968, opened 2016-09-08
Make CollectionSearchBuilder access-level aware

atz:
- Move `configure_blacklight` out one level (from `SelectsCollections` to `CollectionsControllerBehavior`)
- Make `CollectionSearchBuilder` more aware of its inheritance.  Use overriding and initialization instead of adding incessantly to the processor_chain.  We aren't the first ones in this stack to think of sorting by a field or row counts.
- Avoid confusing `discovery_permissions` => `@discovery_perms` accessor, redundant to @discovery_permissions, oddly.
- Cleanup unnecessary overrides. 
- Add a big README for SearchBuilders

### DIFF
--- a/app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
@@ -13,6 +13,9 @@ module CurationConcerns
 
       # This is needed as of BL 3.7
       copy_blacklight_config_from(::CatalogController)
+      configure_blacklight do |config|
+        config.search_builder_class = CurationConcerns::CollectionSearchBuilder
+      end
 
       # Catch permission errors
       rescue_from Hydra::AccessDenied, CanCan::AccessDenied do |exception|
@@ -203,8 +206,8 @@ module CurationConcerns
       alias collection_member_search_builder member_search_builder
       deprecation_deprecate collection_member_search_builder: "use member_search_builder instead"
 
-      def list_search_builder
-        list_search_builder_class.new(self)
+      def list_search_builder(access_level = nil)
+        list_search_builder_class.new(self, access_level)
       end
 
       alias collections_search_builder list_search_builder
--- a/app/controllers/concerns/curation_concerns/selects_collections.rb
+++ b/app/controllers/concerns/curation_concerns/selects_collections.rb
@@ -9,6 +9,8 @@ module CurationConcerns::SelectsCollections
     end
   end
 
+  # @return [Hash{Symbol => Array[Symbol]}] bottom-up map of "what you need" to "what qualifies"
+  # @note i.e., requiring :read access is satisfied by either :read or :edit access
   def access_levels
     { read: [:read, :edit], edit: [:edit] }
   end
@@ -59,8 +61,8 @@ module CurationConcerns::SelectsCollections
   end
 
   def collections_search_builder(access_level = nil)
-    collections_search_builder_class.new(self).tap do |builder|
-      builder.discovery_perms = access_levels[access_level] if access_level
+    collections_search_builder_class.new(self, nil).tap do |builder|
+      builder.discovery_permissions = access_levels[access_level] if access_level
     end
   end
 end
--- a/app/controllers/concerns/curation_concerns/selects_collections.rb
+++ b/app/controllers/concerns/curation_concerns/selects_collections.rb
@@ -61,8 +61,6 @@ module CurationConcerns::SelectsCollections
   end
 
   def collections_search_builder(access_level = nil)
-    collections_search_builder_class.new(self, nil).tap do |builder|
-      builder.discovery_permissions = access_levels[access_level] if access_level
-    end
+    collections_search_builder_class.new(self, access_levels[access_level])
   end
 end
--- a/app/search_builders/curation_concerns/README.md
+++ b/app/search_builders/curation_concerns/README.md
@@ -1,0 +1,69 @@
+# Search Builders
+
+Building searches is core to any Blacklight app, and CurationConcerns is no exception.  
+This directory contains our Search Builders, so named because the design followed a builder pattern, meaning
+that when invoked to set values, methods return the object itself, so that invocations can be chained, like:
+
+```ruby
+builder = Blacklight::SearchBuilder.new(processor_chain, scope)
+            .rows(20)
+            .page(3)
+            .with(q: 'Abraham Lincoln')
+```
+
+However, at this level, many if not most of the additional methods do not follow this pattern.  
+Refer to the `Blacklight::SearchBuilder` class if you want to be certain.  That leads to the next topic...
+
+## Ancestry
+
+Most of the SearchBuilders have `::SearchBuilder` as a parent or ancestor.  `::SearchBuilder` does not exist in any repo: it is generated
+by Blacklight and modified by CurationConcerns.  Others descend from `Blacklight::SearchBuilder`, or various other relatives.  
+
+### ::SearchBuilder
+
+The generated parent class `SearchBuilder` descends from `Blacklight::SearchBuilder`.
+As modified by CurationConcerns' installer, it includes additional modules and overrides.  So if your SearchBuilder has `::SearchBuilder` as a parent class, you are getting:
+- [Blacklight::SearchBuilder](https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/search_builder.rb) grandparent class
+- [Blacklight::Solr::SearchBuilderBehavior](https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/solr/search_builder_behavior.rb) associated methods
+- [Hydra::AccessControlsEnforcement](https://github.com/projecthydra/hydra-head/blob/master/hydra-access-controls/lib/hydra/access_controls_enforcement.rb) module
+  -  [Blacklight::AccessControls::Enforcement](https://github.com/projectblacklight/blacklight-access_controls/blob/master/lib/blacklight/access_controls/enforcement.rb) ancestor of `Hydra::AccessControlsEnforcement`
+- [CurationConcerns::SearchFilters](https://github.com/projecthydra/curation_concerns/blob/master/app/search_builders/curation_concerns/search_filters.rb)  module that itself includes:
+  - [BlacklightAdvancedSearch::AdvancedSearchBuilder](https://github.com/projectblacklight/blacklight_advanced_search/blob/master/lib/blacklight_advanced_search/advanced_search_builder.rb) more magic for compound Boolean queries
+  - [CurationConcerns::FilterByType](https://github.com/projecthydra/curation_concerns/blob/master/app/search_builders/curation_concerns/filter_by_type.rb) Collection vs. Work filtering, specifically the `filter_models` method
+
+This is not a comprehensive list, but it is sufficient to trace some of the complexity of interaction between various layers.
+
+## Development: AKA Doing Something Useful
+
+Note, the `default_processor_chain` defined by `Blacklight::Solr::SearchBuilderBehavior` provides a way to extend functionality, but also many possible points of override (namely method names).  When you need to do something novel and additional, adding to the chain is completely reasonable.  For example:
+
+```ruby
+module MySearchBuilder
+  extend ActiveSupport::Concern
+
+  included do
+    self.default_processor_chain += [:special_filter]
+  end
+
+  def special_filter(solr_parameters)
+    solr_parameters[:fq] << "{!field f=some_field_ssim}#{...}"
+  end
+end
+```
+
+But to the extent that you are overriding (or undoing) something already done, `CurationConcerns::FileSetSearchBuilder` is a better example:
+
+```ruby
+module CurationConcerns
+  class FileSetSearchBuilder < ::SearchBuilder
+    include CurationConcerns::SingleResult
+
+    # This overrides the filter_models in FilterByType
+    def filter_models(solr_parameters)
+      solr_parameters[:fq] << ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: ::FileSet.to_class_uri)
+    end
+  end
+end
+```
+
+There is no point having the other `filter_models` methods apply `:fq`s that we then try to undo or overwrite.  In general, directly overwriting the whole `default_processor_chain` or solr parameters like `:fq` is less flexible than appending constraints sufficient for your use case.  In particular, you might find that you have overwritten components that implement access controls, thereby making your SearchBuilder less useful and less secure.  When in doubt, examine the actual solr queries produced.
--- a/app/search_builders/curation_concerns/collection_search_builder.rb
+++ b/app/search_builders/curation_concerns/collection_search_builder.rb
@@ -30,18 +30,8 @@ module CurationConcerns
       { read: [:read, :edit], edit: [:edit] }
     end
 
-    ## Overrides
-
-    # unprotect lib/blacklight/access_controls/enforcement.rb methods
-    # Remove these when https://github.com/projectblacklight/blacklight-access_controls/pull/23 is merged/released/required
-    def discovery_permissions
-      super
-    end
-
-    def discovery_permissions=(*args)
-      super
-    end
-
+    # @!group Overrides
+    #
     # This overrides the filter_models in FilterByType
     def filter_models(solr_parameters)
       solr_parameters[:fq] ||= []
--- a/app/search_builders/curation_concerns/collection_search_builder.rb
+++ b/app/search_builders/curation_concerns/collection_search_builder.rb
@@ -1,9 +1,6 @@
 module CurationConcerns
   # Our parent class is the generated SearchBuilder descending from Blacklight::SearchBuilder
-  # It includes Blacklight::Solr::SearchBuilderBehavior, Hydra::AccessControlsEnforcement, CurationConcerns::SearchFilters
-  # @see https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/search_builder.rb Blacklight::SearchBuilder parent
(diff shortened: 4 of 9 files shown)


atz:
Some kind of complex dependency issue:
https://travis-ci.org/projecthydra/curation_concerns/jobs/158330882#L195

jcoyne:
Due to browse-everything pinned to google-api-client 0.8.x (breaking changes to 0.9) and then google-api-client doing: https://github.com/google/google-api-ruby-client/commit/fd295849b9409da87db170d60ccd3fed2bcb674c

jcoyne:
One of the working builds has resolved to google-api-client 0.8.6: https://travis-ci.org/projecthydra/curation_concerns/jobs/158306421#L330

jcoyne:
@atz how about removing https://github.com/projecthydra/curation_concerns/pull/968/commits/31b03d3b45db9354a2c44d5b3b9d8493df34e322 from this PR (put it in a new one). See if that fixes the dependency issue.

atz:
I don't think I can do that, since I rely on the recent upstream un-protection of `discovery_permissions=`.

atz:
Here's the api client: https://github.com/google/google-api-ruby-client

app/search_builders/curation_concerns/collection_search_builder.rb
```diff
@@ -1,33 +1,51 @@
...
+  #
   class CollectionSearchBuilder < ::SearchBuilder
     # Defines which search_params_logic should be used when searching for Collections
-    self.default_processor_chain = [:default_solr_parameters, :add_query_to_solr,
-                                    :add_access_controls_to_solr_params, :add_collection_filter, :some_rows, :sort_by_title]
-
-    def some_rows(solr_parameters)
-      solr_parameters[:rows] = '100'
+    def initialize(context, access)
+      @access = access
```
jcoyne: Why are we assigning this variable. Is it used anywhere?

app/search_builders/curation_concerns/collection_search_builder.rb
```diff
@@ -1,33 +1,51 @@
...
+    # @return [String] Solr field name indicating default sort order
+    def sort_field
+      Solrizer.solr_name('title', :sortable)
     end
 
-    attr_writer :discovery_perms
+    ## Overrides
 
+    # unprotect lib/blacklight/access_controls/enforcement.rb methods
+    # Remove these when https://github.com/projectblacklight/blacklight-access_controls/pull/23 is merged/released/required
```
jcoyne: So, now these can be removed?

atz:
Looks like the current master (unreleased) eliminated `activesupport` as a dependency altogether, which would be totally sufficient:
https://github.com/google/google-api-ruby-client/commit/496d795e4c79a0a4c0b5a26ce778957f9f04893c

jcoyne:
Yeah, but that's still the 0.9 API, of which google api is incompatible with.

atz:
@jcoyne not sure what you are saying.  That *is* the `google-api-client`.  What's the incompatibility?  With live google servers or with itself?  Neither of those make sense.  As consumers if we follow the 0.8 -> 0.9 `MIGRATING` doc and use the 0.9 series gems accordingly, what is the issue?

jcoyne:
There will be no issue once we migrate to 0.9, but that's still a hurdle.

atz:
Yeah, but I think between cbeer and I, we have that part done already.

app/search_builders/curation_concerns/collection_search_builder.rb
```diff
@@ -1,33 +1,42 @@
...
-    # Defines which search_params_logic should be used when searching for Collections
-    self.default_processor_chain = [:default_solr_parameters, :add_query_to_solr,
-                                    :add_access_controls_to_solr_params, :add_collection_filter, :some_rows, :sort_by_title]
+    # @param [#blacklight_config, #current_ability] scope the object that has access to #blacklight_config
+    #   From a controller, scope would be `self`.  This argument passed to ::SearchBuilder.new()
+    # @param [Array{String}] permissions defining breadth of search, e.g. 'edit', 'read', 'discover'
+    # @note permissions will otherwise be defaulted by inherited #discovery_permissions
+    def initialize(scope, permissions = nil)
+      @rows = 100
+      @discovery_permissions ||= permissions if permissions && !permissions.empty?
```
jcoyne: how about `permissions.present?` instead of `permissions && !permissions.empty?`

app/search_builders/curation_concerns/README.md
```diff
@@ -0,0 +1,69 @@
...
+that when invoked to set values, methods return the object itself, so that invocations can be chained, like:
+
+```ruby
+builder = Blacklight::SearchBuilder.new(processor_chain, scope)
+            .rows(20)
+            .page(3)
+            .with(q: 'Abraham Lincoln')
+```
+
+However, at this level, many if not most of the additional methods do not follow this pattern.
```
jcoyne: "at this level" doesn't add anything. I think you can omit that.

app/search_builders/curation_concerns/README.md
```diff
@@ -0,0 +1,69 @@
...
+- [Hydra::AccessControlsEnforcement](https://github.com/projecthydra/hydra-head/blob/master/hydra-access-controls/lib/hydra/access_controls_enforcement.rb) module
+  -  [Blacklight::AccessControls::Enforcement](https://github.com/projectblacklight/blacklight-access_controls/blob/master/lib/blacklight/access_controls/enforcement.rb) ancestor of `Hydra::AccessControlsEnforcement`
+- [CurationConcerns::SearchFilters](https://github.com/projecthydra/curation_concerns/blob/master/app/search_builders/curation_concerns/search_filters.rb)  module that itself includes:
+  - [BlacklightAdvancedSearch::AdvancedSearchBuilder](https://github.com/projectblacklight/blacklight_advanced_search/blob/master/lib/blacklight_advanced_search/advanced_search_builder.rb) more magic for compound Boolean queries
+  - [CurationConcerns::FilterByType](https://github.com/projecthydra/curation_concerns/blob/master/app/search_builders/curation_concerns/filter_by_type.rb) Collection vs. Work filtering, specifically the `filter_models` method
+
+Seem like lot?  Well, this is the simplified view.
```
jcoyne: This doesn't add anything to the understanding of someone looking for help in the readme.

app/search_builders/curation_concerns/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+But to the extent that you are overriding (or undoing) something already done, `CurationConcerns::FileSetSearchBuilder` is a better example:
+
+```ruby
+module CurationConcerns
+  class FileSetSearchBuilder < ::SearchBuilder
+    include CurationConcerns::SingleResult
+
+    # This overrides the filter_models in FilterByType
+    def filter_models(solr_parameters)
```
jcoyne: this example will be obsolete when https://github.com/projecthydra/curation_concerns/pull/970 is merged.

app/search_builders/curation_concerns/README.md
```diff
@@ -0,0 +1,69 @@
...
+- [CurationConcerns::SearchFilters](https://github.com/projecthydra/curation_concerns/blob/master/app/search_builders/curation_concerns/search_filters.rb)  module that itself includes:
+  - [BlacklightAdvancedSearch::AdvancedSearchBuilder](https://github.com/projectblacklight/blacklight_advanced_search/blob/master/lib/blacklight_advanced_search/advanced_search_builder.rb) more magic for compound Boolean queries
+  - [CurationConcerns::FilterByType](https://github.com/projecthydra/curation_concerns/blob/master/app/search_builders/curation_concerns/filter_by_type.rb) Collection vs. Work filtering, specifically the `filter_models` method
+
+Seem like lot?  Well, this is the simplified view.
+
+## Development: AKA Doing Something Useful
+
+Note, the `default_processor_chain` defined by `Blacklight::Solr::SearchBuilderBehavior` provides a way to extend functionality, but also many possible points of override (namely method names).  When you need to do something novel and additional, adding to the chain is completely reasonable, like:
```
jcoyne: I'd swap "and" for "but also".  Regarding "namely method names", is there anything else you could override? Can you say "many possible methods to override"?

app/search_builders/curation_concerns/README.md
```diff
@@ -0,0 +1,69 @@
+# Search Builders
+
+Building searches is core to any Blacklight app, and CurationConcerns is no exception.  
+This directory contains our Search Builders, so named because the initial design followed a builder pattern, meaning
```
jcoyne: I'd remove "initial" OR explain what the current design is.

app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
```diff
@@ -15,6 +15,9 @@ module CollectionsControllerBehavior
 
       # This is needed as of BL 3.7
       copy_blacklight_config_from(::CatalogController)
+      configure_blacklight do |config|
+        config.search_builder_class = CurationConcerns::CollectionSearchBuilder
```
jcoyne: I'm pretty sure this isn't necessary because `#index` is calling `collection_search_builder` directly.

app/search_builders/curation_concerns/collection_search_builder.rb
```diff
@@ -1,33 +1,42 @@
...
+    # @return [String] Solr field name indicating default sort order
+    def sort_field
+      'title_si'
     end
 
-    def add_collection_filter(solr_parameters)
+    # @!group Overrides
+
+    # This overrides the filter_models in FilterByType
+    def filter_models(solr_parameters)
```
jcoyne: Rebasing this after #970 is merged in would make this nicer.  Then you just override `#models` (similar to what you've called `model_class_uri`)

app/search_builders/curation_concerns/collection_search_builder.rb
```diff
@@ -1,33 +1,42 @@
...
       solr_parameters[:fq] ||= []
-      solr_parameters[:fq] << ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: ::Collection.to_class_uri)
+      solr_parameters[:fq] << ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: model_class_uri)
     end
 
     # Sort results by title if no query was supplied.
     # This overrides the default 'relevance' sort.
-    def sort_by_title(solr_parameters)
+    def add_sorting_to_solr(solr_parameters)
```
jcoyne: Is it important to have `solr` in the method name? That seems like an implementation detail.

spec/controllers/selects_collections_controller_spec.rb
```diff
@@ -9,8 +9,9 @@ class SelectsCollectionsController < ApplicationController
 describe SelectsCollectionsController, type: :controller do
   describe "#find_collections" do
     it "uses the search builder" do
-      expect(subject.collections_search_builder_class.default_processor_chain).to eq [:default_solr_parameters, :add_query_to_solr, :add_access_controls_to_solr_params, :add_collection_filter, :some_rows, :sort_by_title]
-      expect(CurationConcerns::CollectionSearchBuilder).to receive(:new).with(subject).and_call_original
+      expect(subject.collections_search_builder_class.default_processor_chain)
+        .to include(:default_solr_parameters, :add_query_to_solr, :add_access_controls_to_solr_params)
```
jcoyne: Why aren't you testing for `add_sorting_to_solr`?

app/controllers/concerns/curation_concerns/selects_collections.rb
```diff
@@ -57,8 +51,8 @@ def collections_search_builder_class
   end
 
   def collections_search_builder(access_level = nil)
-    collections_search_builder_class.new(self).tap do |builder|
-      builder.discovery_perms = access_levels[access_level] if access_level
+    collections_search_builder_class.new(self, nil).tap do |builder|
```
jcoyne: Instead of using this tap, why aren't you sending the access level as the second argument? Isn't that what you added it for?

app/search_builders/curation_concerns/README.md
atz: Most of the methods in the Blacklight::SearchBuilder itself do follow the pattern.  But we get a descendent of that class w/ many other behaviors and modules mixed in.  That's the difference.

app/search_builders/curation_concerns/README.md
atz: It added to my understanding.  This was the piece I kept coming back to for reference.

app/controllers/concerns/curation_concerns/selects_collections.rb
atz: Yeah, could do that here.  I did that early on and stopped caring as much about `selects_collections.rb` when you declared it deprecated.

spec/controllers/selects_collections_controller_spec.rb
atz: I do that in `controller_search_builder_spec.rb`.  The class was not tested previously, but now that it is, several tests are more appropriate there.

app/search_builders/curation_concerns/collection_search_builder.rb
atz: Yeah, i guess this is still inside a rails app.  Could do that.

app/search_builders/curation_concerns/README.md
atz: It will still be a valid example to illustrate overriding `filter_models`, it just won't match live code or current `:fq` practices.  Similarly, the previous example is still instructional despite being a fictional class.

app/search_builders/curation_concerns/README.md
jcoyne: The last two sentences?

app/controllers/concerns/curation_concerns/selects_collections.rb
jcoyne: It's deprecated, but the good stuff is copied into the controller, so please do it for the improvement of the controller

spec/controllers/selects_collections_controller_spec.rb
jcoyne: Should we remove this assertion entirely then?

app/search_builders/curation_concerns/collection_search_builder.rb
atz: Yeah, I was tempted to do something similar.  Dislike the name `models` for the method though.

app/search_builders/curation_concerns/collection_search_builder.rb
atz: The name is important because it is included in the default processor chain.  Instead of processing the wrong sort order and then later overwriting it, we can override and process the correct order once.  That's why it is part of `# @!group Overrides`

spec/controllers/selects_collections_controller_spec.rb
atz: I don't check the remaining methods in the other test, so I left them in here.  I'm amenable to an argument for removal, but since SelectsController is deprecated anyway, I'd just as soon not spend additional time on its tests.

app/search_builders/curation_concerns/README.md
atz: Ah, no. That was just to acknowledge the complexity and also that this is not a comprehensive list.  I can rework it to that effect.

app/search_builders/curation_concerns/README.md
atz: "For example" would be fine.  You can override anything: attributes, entire namespaces, even constants.  The point is that `default_processor_chain` is just a list of (Symbol) names, not the methods themselves.

app/search_builders/curation_concerns/README.md
atz: Fair point.

app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
atz: Then it was never necessary since that was the only consumer of the module it moved from?

spec/controllers/selects_collections_controller_spec.rb
atz: Basically, yes, we can remove the whole assertion, but lets just do it when we nuke SelectsController altogether.

app/search_builders/curation_concerns/collection_search_builder.rb
jcoyne: Ah. Okay

spec/controllers/selects_collections_controller_spec.rb
jcoyne: Okay

app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
jcoyne: I think it's vestigial. It was necessary at some point in time, but then `#index` was refactored.

app/search_builders/curation_concerns/collection_search_builder.rb
```diff
@@ -1,33 +1,48 @@
 module CurationConcerns
+  # Our parent class is the generated SearchBuilder descending from Blacklight::SearchBuilder.
+  # Defines search_params_logic used when searching for Collections.
+  # @see https://github.com/projecthydra/curation_concerns/blob/master/app/search_builders/curation_concerns/README.md SearchBuilders README
+  #
   class CollectionSearchBuilder < ::SearchBuilder
-    # Defines which search_params_logic should be used when searching for Collections
-    self.default_processor_chain = [:default_solr_parameters, :add_query_to_solr,
```
jcoyne: This may have been lost during a rebase. We still want this, right?

atz: Removing these is intentional, since we are relying on the common chain built up in `SearchBuilder`.   Instead of needing to re-specify a new chain w/ bespoke methods, we are using (overriding) more of the already defined methods (e.g. `filter_models` instead of `add_collection_filters`).

atz:
Browse-everything is now migrated in master, but no release has been cut yet.

atz:
Closing to pursue in #1105 (and subsequent PRs).

PR #969, opened 2016-09-08
Make search builder classes into class_attributes

jcoyne:

Also updated method names to be more expressive of their purpose.

### DIFF
--- a/app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
@@ -32,9 +32,23 @@ module CurationConcerns
       before_action :authenticate_user!, except: [:show, :index]
       load_and_authorize_resource except: [:index, :show], instance_name: :collection
 
-      class_attribute :presenter_class, :form_class
+      class_attribute :presenter_class,
+                      :form_class,
+                      :single_item_search_builder_class,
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/curation_concerns/collections_controller_behavior.rb
```diff
@@ -32,9 +32,23 @@ module CollectionsControllerBehavior
...
+
+      alias_method :collection_search_builder_class, :single_item_search_builder_class
+      deprecation_deprecate collection_search_builder_class: "use single_item_search_builder_class instead"
+
+      alias_method :collection_member_search_builder_class, :member_search_builder_class
+      deprecation_deprecate collection_member_search_builder_class: "use member_search_builder_class instead"
+
       self.presenter_class = CurationConcerns::CollectionPresenter
       self.form_class = CurationConcerns::Forms::CollectionEditForm
+      # The search builder to find the collection
```
mjgiarlo: Is this comment on line 48 still about the code on line 49? Comment says it's finding the collection but the name of the search builder is oriented around works rather than collections.

jcoyne: Yes, the WorkSearchBuilder finds Collections or Works. The comment is accurate.

PR #970, opened 2016-09-08
Refactor filtering by type

jcoyne:
This gives us a smaller footprint to override.

### DIFF
--- a/app/search_builders/curation_concerns/file_set_search_builder.rb
+++ b/app/search_builders/curation_concerns/file_set_search_builder.rb
@@ -2,9 +2,9 @@ module CurationConcerns
   class FileSetSearchBuilder < ::SearchBuilder
     include CurationConcerns::SingleResult
 
-    # This overrides the filter_models in FilterByType
-    def filter_models(solr_parameters)
-      solr_parameters[:fq] << ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: ::FileSet.to_class_uri)
+    # This overrides the models in FilterByType
+    def models
+      [::FileSet.to_class_uri]
     end
   end
 end
--- a/spec/search_builders/curation_concerns/admin_set_search_builder_spec.rb
+++ b/spec/search_builders/curation_concerns/admin_set_search_builder_spec.rb
@@ -9,7 +9,7 @@ describe CurationConcerns::AdminSetSearchBuilder do
     before { builder.filter_models(solr_params) }
 
     it 'adds AdminSet to query' do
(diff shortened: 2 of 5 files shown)


spec/search_builders/curation_concerns/admin_set_search_builder_spec.rb
```diff
@@ -9,7 +9,7 @@
     before { builder.filter_models(solr_params) }
 
     it 'adds AdminSet to query' do
-      expect(solr_params[:fq].first).to include('{!field f=has_model_ssim}AdminSet')
```
jeremyf: I'm not following why this changed.

jcoyne: These generate functionally equivalent queryies.  "terms" is more correct for our purposes because it's not a fielded search.

jeremyf: My confusion comes from the spec saying "field" and the above `filter_by_type.rb` has "terms" and the "terms" hasn't changed from "field".

jcoyne: This is occurring because I removed unnecessary overrides which were using `field` rather than `terms`

app/search_builders/curation_concerns/file_set_search_builder.rb
```diff
@@ -2,9 +2,9 @@ module CurationConcerns
   class FileSetSearchBuilder < ::SearchBuilder
     include CurationConcerns::SingleResult
 
-    # This overrides the filter_models in FilterByType
-    def filter_models(solr_parameters)
-      solr_parameters[:fq] << ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: ::FileSet.to_class_uri)
+    # This overrides the models in FilterByType
+    def models
```
atz: I'd prefer a method name that made it clear the return is URIs and not model Classes or some other representation of models.  This is a complaint for FilterByType, I guess, but that method doesn't exist in my branch, so it must only have been added recently.

cbeer: 👍  `#model_uris`?

jcoyne: The problem is that `to_class_uri` doesn't return a uri. It simply returns a model name.  This is a vestige from the Fedora 3 days.

jcoyne: See https://github.com/projecthydra/active_fedora/pull/1148/files

PR #989, opened 2016-09-14
Consolidate CollectionsHelper and CurationConcerns::CollectionsHelper

jcoyne:


mjgiarlo:
@jcoyne I dig the consolidation, but does removing the behavior make it more difficult for upstream consumers to override default CC behavior?

jcoyne:
Not at all, since the module is namespaced.

mjgiarlo:
@jcoyne The pattern before was you could create your own `CollectionsHelper` in the global namespace, and include the behavior in the class, and then you could override, say, one method. Now if you want to do that, you create `app/helpers/curation_concerns/collections_helper.rb`, copy/paste the code from CC, and tweak the one method you want to behave differently?

jcoyne:
@mjgiarlo you can do the same but with the MainAppHelper https://github.com/projecthydra/curation_concerns/blob/8993271bc5eca21b9aeffc885482c952f5d3586e/app/helpers/curation_concerns/main_app_helpers.rb#L10

jcoyne:
@mjgiarlo or stick it right in CurationConcernsHelper: https://github.com/projecthydra/curation_concerns/blob/8f190ff6e93ffb1a95574727c7865d4dc9d0f682/lib/generators/curation_concerns/templates/curation_concerns_helper.rb#L1

mjgiarlo:
@jcoyne thanks!

PR #992, opened 2016-09-16
Add to collections should only show editable collections

jcoyne:


spec/controllers/curation_concerns/collections_controller_spec.rb
```diff
@@ -245,11 +245,20 @@
...
+      create(:collection, :public)
+      sign_in user
+    end
 
-    it 'does not show flash' do
+    it 'is successful' do
       get :edit, params: { id: collection }
       expect(flash[:notice]).to be_nil
+      # a list of collections I can add items to:
+      expect(assigns[:user_collections].map(&:id)).to match_array [collection.id,
```
cam156: @jcoyne wasn't collection not supposed to show up?

jcoyne: The collection created on 252 does not show up. The two that match are on line 25 and on line 248.

jcoyne: The collection created on 252 does not show up. The two that match are on line 25 and on line 248.

jcoyne:
@cam156 added the test in the second commit that moves the CollectionsService from Sufia to CC.

app/services/curation_concerns/collections_service.rb
```diff
@@ -0,0 +1,28 @@
...
+    def search_results(access)
+      builder = list_search_builder(access)
+      response = context.repository.search(builder)
+      response.documents
+    end
+
+    private
+
+      def list_search_builder(access)
+        list_search_builder_class.new(context).tap do |builder|
```
atz: This section we would hope to replace when the upstream hydra-head blocked by browse_everything is finally unblocked.  The whole method can then just go away and be replaced by a constructor call.

mjgiarlo:
I believe all concerns expressed on this PR have been addressed. Merging!

PR #993, opened 2016-09-20
Use Rspec 3.5 or later

awead:
Reported by @jeremyf. Rspec will fail at 3.4.


jeremyf:
We are relying on `config.shared_context_metadata_behavior=` which was added https://github.com/rspec/rspec-core/compare/v3.5.0.beta3...v3.5.0.beta4

PR #1036, opened 2016-09-23
Adding a module that includes items for a controller to act as an admin controller

cam156:
A few things need to be included for another controller to successfully act as an admin controller.

This PR creates a module for including those things.


app/controllers/concerns/curation_concerns/act_as_admin_controller.rb
```diff
@@ -0,0 +1,12 @@
+module CurationConcerns
+  module ActAsAdminController
```
jcoyne: Please add module documentation including an example usage.

cam156: @jcoyne Added Some comments.

PR #1044, opened 2016-09-30
Remove blacklight-advanced-search

jcoyne:
This doesn't seem to be doing anything useful.


jpstroop:
:+1: No use case for it here.

PR #1046, opened 2016-10-06
Make assign_visibility work with symbol hash keys

mkorcy:
`create_metadata` may ignore file_set_params

I would expect to be able to set the visibility of a file_set as followes
```ruby
ds_actor = CurationConcerns::Actors::FileSetActor.new(file_set, user)
ds_actor.create_metadata(obj, {visibility: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE})
```

However, because the hash keys are compared as strings in the private method `assign_visibility`, this example fails to match and thus visibility is always inherited from the parent rather than assigned.  Initially, I considered changing the comparison to just compare symbols, `%i{}` but there are other tests that fail because there are other occasions where hash keys are passed as Strings.

* allow `assign_visibility` and thus `create_metadata` to accept both strings and symbols as hash keys.

### DIFF
--- a/app/actors/curation_concerns/actors/file_set_actor.rb
+++ b/app/actors/curation_concerns/actors/file_set_actor.rb
@@ -125,7 +125,7 @@ module CurationConcerns
         end
 
         def assign_visibility?(file_set_params = {})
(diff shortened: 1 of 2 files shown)


escowles:
👍

PR #1048, opened 2016-10-10
Handle nil assignment of file_set_params

mkorcy:
refs #1046

Handle nil file_set_params in assign_visibility

Addresses issue @jcoyne raised on previous pull request referenced above.


jcoyne:
Thanks for following up on this @mkorcy

PR #1053, opened 2016-10-12
Added bin/rails which should be present in an engine

jcoyne:
This allows us to generate migrations and such.


jcoyne:
Looks like this fights against engine_cart.

PR #1058, opened 2016-10-12
Adding LicenseService dependency injection

jeremyf:
Adding LicenseService dependency injection

Building on the request of @pgwillia on #1047, I'm exposing a means of
leveraging a different license service class. This is useful as some
places allow the user to free-form fill out a license; While I'm not
sure of the general usage, this is a demonstration of how it might be
done.

### DIFF
--- a/lib/curation_concerns/configuration.rb
+++ b/lib/curation_concerns/configuration.rb
@@ -221,6 +221,18 @@ module CurationConcerns
       registered_curation_concern_types.map(&:constantize)
     end
 
+    # A configuration point for changing the behavior of the license service.
+    #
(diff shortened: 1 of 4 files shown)


lib/curation_concerns/configuration.rb
```diff
@@ -221,6 +221,18 @@ def curation_concerns
       registered_curation_concern_types.map(&:constantize)
     end
 
+    # A configuration point for changing the behavior of the license service.
+    #
+    # @!attribute [w] import_export_jar_file_path
```
jcoyne: This documentation doesn't seem correct.

PR #1060, opened 2016-10-14
Add Sipity Workflow to CurationConcerns

jcoyne:


jpstroop:
I wonder if squashing this down into a few logical chunks (something less than 124) might make review a little easier.  @escowles also suggested it might be a good topic for next week's tech call.

mjgiarlo:
@tpendragon Would it be helpful if @jcoyne did what @jpstroop recommended above, or meh?

jcoyne:
@tpendragon *POKE* have you had a chance to look at this?

tpendragon:
@jcoyne Sorry, still looking at it. I don't mean to be holding everyone up, local priorities have been taking up my time.

app/services/curation_concerns/workflow/workflow_schema.rb
```diff
@@ -0,0 +1,41 @@
...
+        optional(:description).filled(:str?) # Sipity::Workflow#description
+        required(:actions).each do
+          required(:name).filled(:str?) # Sipity::WorkflowAction#name
+          required(:from_states).each do
+            required(:names) { array? { each(:str?) } } # Sipity::WorkflowState#name
+            required(:roles) { array? { each(:str?) } } # Sipity::Role#name
+          end
+          optional(:transition_to).filled(:str?) # Sipity::WorkflowState#name
+          optional(:notifications).each do
+            required(:name).value(format?: /\A[a-z|_]+\Z/i) # Sipity::Notification#name
```
mjgiarlo: @blancoj In order to resolve notification names onto namespaced classes, we will need to change the regex here to:

```ruby
/\A[a-z|_\/]+\Z/i
```

Thoughts, @jcoyne @jeremyf ?

jeremyf: Please include a namespace test

app/views/curation_concerns/base/_workflow_actions.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        <div class="form-group">
+          <label for="workflow_action_comment">Review comment:</label>
+          <textarea class="form-control" name="workflow_action[comment]" id="workflow_action_comment"></textarea>
+        </div>
+
+        <input class="btn btn-primary" type="submit" value="Submit">
+
+        <h4>Previous Comments</h4>
+        <dl>
+          <% presenter.workflow.comments.each do |comment| %>
```
jeremyf: Would prefer to see `presenter.workflow_comments.each`. This can be quickly achieved in the presenter by adding a delegate method.

```ruby
delegate :comments, to: :workflow, prefix: 'workflow'
```

tpendragon:
In the workflow roles controller, the "roles" should definitely be a multi-select box.

tpendragon:
Besides a couple tiny comments above, the code looks fine, and it works great. It looks like the workflow generator can be run multiple times without any issues, too.

However, a big thing this is missing for us is programmatic role assignment - some users of our application have roles as a side effect of some properties they have (in our case, all users can "flag"). Having it all in the database with no interface between the parts means I can't do that, at least as far as I can tell. The best I can do is a bunch of after_save hooks on users which auto-build roles. I'm concerned that the database structure is the interface. Thoughts?

jcoyne:
@tpendragon I will work on supporting Groups as Agents this week.  This should allow you to grant "flag" to any "registered" user, without callbacks.

tpendragon:
@jcoyne I see you did that - with this, how does one assign a group to roles?

jcoyne:
@tpendragon 
```ruby
group = CurationConcerns::Group.new('registered')
group_agent = group.to_sipity_agent
PermissionGenerator.call(roles: 'reviewing', workflow: sipity_workflow, agents: group_agent)
```

PR #1063, opened 2016-10-19
Disable Turbolinks for File Downloads

### LINKED ISSUE #1050: "Download this File" fails within existing browser windows
Affects at least 0.12.0 through 5ed966284a2123a50e40414e32f1ab64a7738d5b.

cc @val99erie 

1. `rake engine_cart:generate`
2. `cd .internal_test_app && bundle exec fcrepo_wrapper`
3. `cd .internal_test_app && bundle exec solr_wrapper`
4. `cd .internal_test_app && bin/rails s -b 0.0.0.0`
5. Sign up, create a new generic work, add a file to it.
6. Go to the FileSet show page (`http://localhost:3000/concern/file_sets/<id>`), and click "Download this File"

### Expected behavior

The PDF viewer displays the file, or the file is downloaded if there's no in-browser PDF viewer.

### Actual behavior

The browser attempts to render the PDF as text:
screen shot 2016-10-10 at 10 29 05 am

When the file is an MP3, the browser freezes.

Opening the link in a new tab or window results in the correct behavior, however.

no-reply:
Fixes #1050.

Turbolinks causes xhr requests on local links. For file downloads, this causes browsers to render the contents as text, instead of playing media or displaying the PDF in a viewer. This disables it for the download links.

Related proposed changes:

- Replace href target "_new" with "_blank". Since `_new` is not a valid browsing context name in HTML, clients may ignore it. Before disabling turbolinks explicitly, this caused the bad rendering behavior described above in those clients. (44e47e2)
- Make download `_blank` behavior consistent across file download links. (5f8fe1b)
- Update existing `data-no-turbolink` calls to the Turbolinks 5 syntax. (2e9a794)

### DIFF
--- a/app/views/curation_concerns/file_sets/_show_actions.html.erb
+++ b/app/views/curation_concerns/file_sets/_show_actions.html.erb
@@ -1,5 +1,5 @@
   <div class="form-actions">
-    <%= link_to "Download this File", main_app.download_path(presenter), class: 'btn btn-default' %>
+    <%= link_to "Download this File", main_app.download_path(presenter), class: 'btn btn-default', data: { turbolinks: false } %>
     <% if can? :edit, presenter.id %>
       <%= link_to "Edit this File", edit_polymorphic_path([main_app, presenter]), class: 'btn btn-default'  %>
     <% end %>
--- a/app/views/curation_concerns/file_sets/_show_actions.html.erb
+++ b/app/views/curation_concerns/file_sets/_show_actions.html.erb
@@ -1,5 +1,5 @@
   <div class="form-actions">
-    <%= link_to "Download this File", main_app.download_path(presenter), class: 'btn btn-default', data: { turbolinks: false } %>
+    <%= link_to "Download this File", main_app.download_path(presenter), class: 'btn btn-default', data: { turbolinks: false }, target: :_blank %>
     <% if can? :edit, presenter.id %>
       <%= link_to "Edit this File", edit_polymorphic_path([main_app, presenter]), class: 'btn btn-default'  %>
     <% end %>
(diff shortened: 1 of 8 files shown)


no-reply:
Seems like `_new` has been in use from the start: https://github.com/projecthydra/curation_concerns/commit/9d8ebf17b761946ab74576def080504d490d5737

dunn:
This is the link responsible for the behavior in #1050: https://github.com/projecthydra/curation_concerns/blob/master/app/views/curation_concerns/file_sets/_show_actions.html.erb#L2

no-reply:
@dunn thanks! I guess I had misunderstood about this being specific to the "Download This File" link. 052bf28 fixes it for that link.

app/views/curation_concerns/file_sets/_show_actions.html.erb
```diff
@@ -1,5 +1,5 @@
   <div class="form-actions">
-    <%= link_to "Download this File", main_app.download_path(presenter), class: 'btn btn-default' %>
+    <%= link_to "Download this File", main_app.download_path(presenter), class: 'btn btn-default', target: :_blank %>
```
jcoyne: Why does this need to have a _blank target? Can't it open in the same window? I'm unclear as to how this is related to the other commit in the PR.

dunn: Maybe it should be part of a separate PR, but that's the workaround for the issue in #1050.

no-reply: I should say, I'm completely open to another fix and removing the `target` attribute from all of them. This fix is consistent and more convenient.

If this gets merged and #1050 closed, we could consider opening another issue to sort out the xhr issue.

no-reply: The alternate fix is `data: { turbolinks: false }` for the download links. I'm going to rework the PR to address that issue first.

no-reply:
Rebased and force pushed to more clearly handle the various problems at play, here. I'm happy to break this into multiple PRs, if that's best.

PR #1065, opened 2016-10-20
No turbolinks downloads backport

no-reply:
Backports #1063, addressing #1050 for this branch.

> Turbolinks causes xhr requests on local links. For file downloads, this causes browsers to render the contents as text, instead of playing media or displaying the PDF in a viewer. This disables it for the download links.
>Related proposed changes:
> - Replace href target "_new" with "_blank". Since _new is not a valid browsing context name in HTML, clients may ignore it. Before disabling turbolinks explicitly, this caused the bad rendering behavior described above in those clients. (44e47e2)
> - Make download _blank behavior consistent across file download links. (5f8fe1b)
> - Update existing data-no-turbolink calls to the Turbolinks 5 syntax. (2e9a794)

I considered only backporting the first commit, but decided it would be easier to make additional changes to these files if they were kept in sync; and that any (very small) risk of unexpected behavior changes on the 1.6-stable release series was outweighed by this consideration.


awead:
LGTM, and 😍  the commit messages!

PR #1071, opened 2016-10-28
Adds inspect_work endpoint for registered work types

### LINKED ISSUE #1026: Create a debug view for an object's Workflow
In Sipity we have a debug page that allows users to see the state of the Sipity workflow:
- An object's Sipity::Workflow
- An object's Sipity::WorkflowState
- The list of Sipity::Roles that are assigned to the object, and how agents are associated with those roles (i.e. at the Workflow level or the Sipity::Entity level)

Example https://github.com/ndlib/sipity/blob/master/artifacts/for-sufia/sipity-ETD-Debug-page.png

randalldfloyd:
Fixes #1026 

This change adds the ability to request a raw debug view of the underlying workflow states for a work. Such a view bypasses logic interpreting workflows in other views, which could be misleading if troubleshooting a related feature. It is also generally useful to see the state of things without resorting to browsing objects in the console.

It is also more generally useful than for just workflows, as any aspect of a work can be tapped into for raw inspection by following the patterns in the presenter.

The name `inspect_work` was chosen over `debug` since you might not be trying to fix a problem, and `inspect` was avoided to differentiate from the Ruby object `inspect` methods. 

- A new route to `concern/(work_type)/:id/inspect_work` for works based on registered work types
- Corresponding controller action, presenter, and view to handle the request
- A new key in i18n in default locale for basic translations in the view

### DIFF
--- a/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
+++ b/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
@@ -19,7 +19,7 @@ module CurationConcerns
 
     module ClassMethods
       def curation_concern_type=(curation_concern_type)
-        load_and_authorize_resource class: curation_concern_type, instance_name: :curation_concern, except: [:show, :file_manager]
+        load_and_authorize_resource class: curation_concern_type, instance_name: :curation_concern, except: [:show, :file_manager, :inspect_work]
         self._curation_concern_type = curation_concern_type
       end
 
@@ -102,6 +102,11 @@ module CurationConcerns
     end
 
     def file_manager
+      presenter
+    end
+
+    def inspect_work
+      raise Hydra::AccessDenied unless current_ability.admin?
       presenter
     end
 
--- a/app/presenters/curation_concerns/inspect_work_presenter.rb
+++ b/app/presenters/curation_concerns/inspect_work_presenter.rb
@@ -1,0 +1,52 @@
+module CurationConcerns
+  class InspectWorkPresenter
+    def initialize(solr_document, current_ability)
+      @solr_document = solr_document
+      @current_ability = current_ability
+    end
+    attr_reader :solr_document, :current_ability
+
+    def workflow
+      return { comments: [], roles: [] } unless sipity_entity
+      {
+        entity_id: sipity_entity.id,
+        proxy_for: sipity_entity.proxy_for_global_id,
+        workflow_id: sipity_entity.workflow_id,
+        workflow_name: sipity_entity.workflow_name,
+        state_id: sipity_entity.workflow_state_id,
+        state_name: sipity_entity.workflow_state_name,
+        comments: workflow_comments,
+        roles: sipity_entity_roles
+      }
+    end
+
+    def solr
+      @solr_document.inspect
+    end
+
+    private
+
+      def sipity_entity
+        @sipity_entity ||= PowerConverter.convert(solr_document, to: :sipity_entity)
+      end
+
+      def sipity_entity_roles
+        roles = CurationConcerns::Workflow::PermissionQuery.scope_roles_associated_with_the_given_entity(entity: solr_document)
+        roles.map do |role|
+          { id: role.id, name: role.name, description: role.description, users: sipity_entity_role_users(role) }
+        end
+      end
+
+      def sipity_entity_role_users(role)
+        users = CurationConcerns::Workflow::PermissionQuery.scope_users_for_entity_and_roles(entity: sipity_entity, roles: role)
+        users.map(&:user_key)
+      end
+
+      def workflow_comments
+        return [] unless sipity_entity && sipity_entity.comments.count > 0
+        sipity_entity.comments.map do |comment|
+          { comment: comment.comment, created_at: comment.created_at }
+        end
+      end
+  end
+end
--- a/app/views/curation_concerns/base/inspect_work.html.erb
+++ b/app/views/curation_concerns/base/inspect_work.html.erb
@@ -1,0 +1,37 @@
+<h1><%= t("inspect_work.link_text") %></h1>
+<ul class="breadcrumb">
+  <li>
+    Back to <%= link_to @presenter.to_s, [main_app, @presenter] %>
+  </li>
+</ul>
+
+<h3>Workflow</h3>
+<dl>
+  <% workflow = @presenter.inspect_work.workflow %>
+  <dt>Object Name</dt>
+  <dd><%= @presenter.to_s %></dd>
+  <dt>Processing Entity ID</dt>
+  <dd>Entity ID: <%= workflow[:entity_id] %> (Proxy for <%= workflow[:proxy_for] %>)</dd>
+  <dt>Workflow</dt>
+  <dd>Name: <%= workflow[:workflow_name] %> (ID=<%= workflow[:workflow_id] %>)</dd>
+  <dt>State</dt>
+  <dd>State Name: <%= workflow[:state_name] %> (ID=<%= workflow[:state_id] %>)</dd>
+  <dt>Workflow Comments</dt>
+  <dd>
+    <% workflow[:comments].each do |comment| %>
+        <dd><%= comment[:created_at] %>, <%= comment[:comment] %></dd>
+    <% end %>
+  </dd>
+  <dt>Roles</dt>
+  <dd>
+    <% workflow[:roles].each do |role| %>
+      <dd><%= role[:name] %> (ID=<%= role[:id] %>) <%= role[:description] %> Users: <%= role[:users] %></dd>
+  <% end %>
+  </dd>
+</dl>
+
+<h3>Persistence</h3>
+<dl>
+  <dt>Solr</dt>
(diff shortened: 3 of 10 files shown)


randalldfloyd:
A few caveats on this PR: I didn't have time to figure out how to implement everything we want to inspect for workflows so a few TODO's are marked in the presenter (i.e. displaying associated actionable roles and their associated actors for a `:sipity_entity`.)  I also am not adding or checking for additional ability beyond what the user currently has for the work, which is probably wrong. Should it only be exposed to admin or workflow roles?

This is good first pass that introduces all the pieces but has lots of room for extending or improving.

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+    private
+
+      def sipity_entity
+        @sipity_entity ||= PowerConverter.convert(solr_document, to: :sipity_entity)
+      rescue PowerConverter::ConversionError
+        nil
+      end
+
+      def sipity_role
+        # TODO: How to use PowerConverter to get info on roles for a sipity_entity
```
jcoyne: https://github.com/projecthydra/curation_concerns/blob/workflow/app/services/curation_concerns/workflow/permission_query.rb#L119

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+        comments = []
+        sipity_entity.comments.each do |comment|
+          comments << { comment: comment.comment, created_at: comment.created_at }
+        end
+        comments
+      end
+
+      def work
+        # This presenter just has a solr_document but we might want to inspect the work object
+        # TODO: Is there a better way at this point to get back to the actual work object
```
jcoyne: There is no better way, but you should just be able to inspect the solr_doc. I can't see any reason to load the work from fedora (it will be slower)

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+          state_id: sipity_entity.workflow_state_id,
+          state_name: sipity_entity.workflow_state_name,
+          comments: workflow_comments,
+          roles: sipity_role # TODO: How to get related roles and actors for the sipity_entity
+        }
+      end
+    end
+
+    def solr
+      work.to_solr if work
```
jcoyne: This causes a solr_document -> fedora_cocument -> solr_document.  Why would we want to do that?

app/views/curation_concerns/base/inspect_work.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <dd><%= comment[:created_at] %>, <%= comment[:comment] %></dd>
+  <% end %>
+  <dt>Roles</dt>
+  <dd><%= workflow[:roles] %></dd>
+</dl>
+
+<h3>Storage</h3>
+<dl>
+  <dt>Work persistence</dt>
+  <dd><%= @presenter.inspect_work.storage %></dd>
```
jcoyne: I would just leave this part out, it's slow and unlikely to give you anything useful that the solr version wouldn't

app/presenters/curation_concerns/inspect_work_presenter.rb
randalldfloyd: @jcoyne Here was my thinking, though my implementation is certainly lacking: When I'm running down a problem (in general, not just workflow related), I'm usually in the console asking 3 things: what does the object look like in Fedora, what is it _supposed_ to look like if I were to solrize it right now, and what does it really look like in Solr right now? When there's a problem, those things may not be as expected and can show where the problem lies. This view would shortcut that process in a single GET but I'm fine to just avoid for now if it isn't clean or useful.

app/presenters/curation_concerns/inspect_work_presenter.rb
jcoyne: `solr_document.inspect`

app/presenters/curation_concerns/inspect_work_presenter.rb
jcoyne: I'm not sure we want to tackle that right now.  We should emphasize the workflow.

randalldfloyd:
Requested changes addressed and pushed.

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,66 @@
...
+        @roles = CurationConcerns::Workflow::PermissionQuery.scope_roles_associated_with_the_given_entity(entity: solr_document)
+        roles = []
+        @roles.each do |role|
+          roles << { id: role.id, name: role.name, description: role.description, users: sipity_entity_role_users(role) }
+        end
+        roles
+      end
+
+      def sipity_entity_role_users(role)
+        @users = CurationConcerns::Workflow::PermissionQuery.scope_users_for_entity_and_roles(entity: sipity_entity, roles: role)
```
jcoyne: Why is this an instance variable?

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,66 @@
...
+        users = []
```
jcoyne: These 5 lines could be changed to be: `@users.map { |user| user.user_key }` It's preferable to use `user_key`, because that field can be customized although it is `email` by default. Some institutions use `username`

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,66 @@
...
+        @users.each do |user|
+          users << user.email
+        end
+        users
+      end
+
+      def workflow_comments
+        return [] unless sipity_entity && sipity_entity.comments.count > 0
+        comments = []
```
jcoyne: again, here it's idiomatic to use `map`:
```ruby
sipity_entity.comments.map do |comment|
  { comment: comment.comment, created_at: comment.created_at }
end
```

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,66 @@
...
+    private
+
+      def sipity_entity
+        @sipity_entity ||= PowerConverter.convert(solr_document, to: :sipity_entity)
+      rescue PowerConverter::ConversionError
+        nil
+      end
+
+      def sipity_entity_roles
+        @roles = CurationConcerns::Workflow::PermissionQuery.scope_roles_associated_with_the_given_entity(entity: solr_document)
```
jcoyne: Why is this an instance variable?

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,66 @@
...
+        roles = []
+        @roles.each do |role|
```
jcoyne: Use `map` here as well (see below)

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def solr
+      @solr_document.inspect
+    end
+
+    private
+
+      def sipity_entity
+        @sipity_entity ||= PowerConverter.convert(solr_document, to: :sipity_entity)
+      rescue PowerConverter::ConversionError
```
jcoyne: You should not catch this error.  If you do, you'll end up with null pointer errors, which are much harder to diagnose than the underlying `ConversionError`

app/presenters/curation_concerns/inspect_work_presenter.rb
```diff
@@ -0,0 +1,66 @@
...
+  class InspectWorkPresenter
+    def initialize(solr_document, current_ability)
+      @solr_document = solr_document
+      @current_ability = current_ability
+    end
+    attr_reader :solr_document, :current_ability
+
+    def workflow
+      workflow_resp = { comments: [], roles: [] }
+      if sipity_entity
```
jcoyne: Use a guard clause here instead.  This means `return workflow_resp if sipity_entity`  then the next line you can just define the Hash (no need to assign it to `workflow_resp`

jcoyne:
Can you point your branch against the `master` branch now that `workflow` has been merged?

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -105,6 +105,11 @@ def file_manager
       presenter
     end
 
+    def inspect_work
+      # TODO: How to hide this action behind admin role. It currently depends on user ability for work
```
jcoyne: This comment is not clear to me.  Can you explain more about what is needed?

randalldfloyd: @jcoyne The TODO above (constraining the inspect_work view to the admin role) has been completed and pushed. The controller test was changed accordingly to expect admin ability.

PR #1080, opened 2016-11-05
Render the unavailable template if user cannot act on a work in current workflow state

luisgreg99:
Fixes projecthydra/sufia#2727

### DIFF
--- a/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
+++ b/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
@@ -15,6 +15,8 @@ module CurationConcerns
       self.work_form_service = WorkFormService
       attr_accessor :curation_concern
       helper_method :curation_concern, :contextual_path
+
+      rescue_from WorkflowAuthorizationException, with: :render_unavailable
     end
 
     module ClassMethods
@@ -198,10 +200,40 @@ module CurationConcerns
         search_result_document(params)
       end
 
+      # Only returns unsuppressed documents the user has read access to
       def search_result_document(search_params)
         _, document_list = search_results(search_params)
-        raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
-        document_list.first
+        return document_list.first unless document_list.empty?
+        document_not_found!
+      end
+
+      def document_not_found!
+        doc = SolrDocument.find(params[:id])
+        raise WorkflowAuthorizationException if doc.suppressed?
+        raise CanCan::AccessDenied.new(nil, :show)
+      end
+
+      def render_unavailable
+        message = I18n.t("curation_concerns.workflow.unauthorized")
+        respond_to do |wants|
+          wants.html do
+            flash[:notice] = message
+            render 'unavailable', status: :unauthorized
+          end
+          wants.json do
+            render plain: message, status: :unauthorized
+          end
+          additional_response_formats(wants)
+          wants.ttl do
+            render plain: message, status: :unauthorized
+          end
+          wants.jsonld do
+            render plain: message, status: :unauthorized
+          end
+          wants.nt do
+            render plain: message, status: :unauthorized
+          end
+        end
       end
   end
 end
--- a/app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
+++ b/app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
@@ -9,8 +9,10 @@ module CurationConcerns
     extend ActiveSupport::Concern
     include CurationConcerns::FilterSuppressed
 
+    # Skip the filter if the current user is permitted to take
+    # workflow actions on the work corresponding to the SolrDocument
+    # with id = `blacklight_params[:id]`
     def only_active_works(solr_parameters)
-      # Do not filter works if current user is permitted to take workflow actions
       return if user_has_active_workflow_role?
       super
     end
--- a/spec/controllers/curation_concerns/generic_works_controller_spec.rb
+++ b/spec/controllers/curation_concerns/generic_works_controller_spec.rb
@@ -12,6 +12,18 @@ describe CurationConcerns::GenericWorksController do
     sign_in user
   end
 
+  describe 'integration test for suppressed documents' do
+    let(:work) do
+      create(:work, :public, state: Vocab::FedoraResourceStatus.inactive)
+    end
+    it 'renders the unavailable message because it is in workflow' do
+      get :show, params: { id: work }
+      expect(response.code).to eq '401'
+      expect(response).to render_template(:unavailable)
+      expect(flash[:notice]).to eq 'The work is not currently available because it has not yet completed the approval process'
+    end
+  end
+
   describe '#show' do
     context 'my own private work' do
       let(:work) { create(:private_generic_work, user: user) }
@@ -59,7 +71,7 @@ describe CurationConcerns::GenericWorksController do
           allow(presenter).to receive(:export_as_ttl).and_return("ttl graph")
         end
 
-        it 'renders an turtle file' do
+        it 'renders a turtle file' do
           get :show, params: { id: '99999999', format: :ttl }
           expect(response).to be_successful
           expect(response.body).to eq "ttl graph"
@@ -74,6 +86,35 @@ describe CurationConcerns::GenericWorksController do
       it 'someone elses private work should show me the page' do
         get :show, params: { id: work }
         expect(response).to be_success
+      end
+    end
+
+    context 'with work still in workflow' do
+      before do
+        allow(controller).to receive(:search_results).and_return([nil, document_list])
+      end
+      context 'with a user lacking workflow permission' do
+        before do
+          allow(SolrDocument).to receive(:find).and_return(document)
+        end
+        let(:document_list) { [] }
+        let(:document) { instance_double(SolrDocument, suppressed?: true) }
+        it 'shows the unauthorized message' do
+          get :show, params: { id: '99999' }
+          expect(response.code).to eq '401'
+          expect(response).to render_template(:unavailable)
+          expect(flash[:notice]).to eq 'The work is not currently available because it has not yet completed the approval process'
+        end
+      end
+      context 'with a user granted workflow permission' do
+        let(:document_list) { [document] }
+        let(:document) { instance_double(SolrDocument) }
+        it 'renders without the unauthorized message' do
+          get :show, params: { id: '88888' }
+          expect(response.code).to eq '200'
+          expect(response).to render_template(:show)
+          expect(flash[:notice]).to be_nil
+        end
       end
     end
 
--- a/app/views/curation_concerns/base/unavailable.html.erb
+++ b/app/views/curation_concerns/base/unavailable.html.erb
@@ -1,0 +1,10 @@
+<h1><%= @presenter %> </h1>
+<span class="state state-<%= @presenter.workflow.state %>"><%= @presenter.workflow.state_label %></span>
+<% if @parent_presenter %>
+    <ul class="breadcrumb">
+        <li><%= link_to @parent_presenter, polymorphic_path([main_app, @parent_presenter]) %></li>
+        <li class="active"><%= @presenter.human_readable_type %></li>
+    </ul>
+<% else %>
+    <span class="human_readable_type">(<%= @presenter.human_readable_type %>)</span>
+<% end %>
--- a/config/locales/curation_concerns.en.yml
+++ b/config/locales/curation_concerns.en.yml
@@ -181,6 +181,7 @@ en:
     workflow:
       default:
         deposit: "Deposit"
+      unauthorized: "The work is not currently available because it has not yet completed the approval process"
   blacklight:
     search:
       fields:
--- a/lib/curation_concerns/workflow_authorization_exception.rb
+++ b/lib/curation_concerns/workflow_authorization_exception.rb
@@ -1,0 +1,4 @@
+module CurationConcerns
+  class WorkflowAuthorizationException < StandardError
+  end
+end
--- a/spec/views/curation_concerns/base/unavailable.html.erb_spec.rb
+++ b/spec/views/curation_concerns/base/unavailable.html.erb_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+
+describe 'curation_concerns/base/unavailable.html.erb', type: :view do
+  let(:model) do
+    double('model',
+           persisted?: true,
+           to_param: '123',
+           model_name: GenericWork.model_name)
+  end
+  let(:workflow) do
+    double('workflow', state: 'deposited', state_label: 'really deposited')
+  end
+  let(:presenter) do
+    double('presenter',
+           to_s: 'super cool',
+           workflow: workflow,
+           human_readable_type: 'Generic Work')
+  end
+  let(:parent_presenter) do
+    double('parent_presenter',
+           to_s: 'parental remark',
+           to_model: model,
+           human_readable_type: 'Foo Bar')
+  end
+  before do
+    assign(:presenter, presenter)
+    assign(:parent_presenter, parent_presenter)
+    stub_template 'shared/_brand_bar.html.erb' => ''
+    stub_template 'shared/_title_bar.html.erb' => ''
+    flash[:notice] = I18n.t("curation_concerns.workflow.unauthorized")
+    render template: 'curation_concerns/base/unavailable.html.erb', layout: 'layouts/curation_concerns'
+  end
+  it "renders with the flash message" do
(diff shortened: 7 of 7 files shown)


app/controllers/concerns/curation_concerns/application_controller_behavior.rb
```diff
@@ -34,6 +34,7 @@ def deny_access(exception)
         respond_to do |wants|
           wants.html do
             if [:show, :edit, :create, :update, :destroy].include? exception.action
+              flash[:alert] = exception.message
```
jcoyne: I'm not sure we want to expose this as it's not an appropriate message for a non-programmer.

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -197,7 +197,11 @@ def curation_concern_from_search_results
 
     def search_result_document(search_params)
       _, document_list = search_results(search_params)
-      raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
+      if (_curation_concern_type.find(params[:id]).suppressed?)
```
jcoyne: This is going to be too slow.  We want to query this from Solr, not from Fedora.

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -197,7 +197,11 @@ def curation_concern_from_search_results
...
+        raise CanCan::AccessDenied.new("The work is not currently viewable because it has not yet completed the approval process", :show) if document_list.empty?
```
jcoyne: I18n

app/forms/curation_concerns/forms/workflow_action_form.rb
```diff
@@ -33,6 +33,7 @@ def save
         handle_sipity_notifications(comment: comment)
         handle_additional_sipity_workflow_action_processing(comment: comment)
         work.update_index # So that the new actions and state are written into solr.
+        work.save
```
jcoyne: why are we doing this save?

app/models/curation_concerns/state_workflow.rb
```diff
@@ -7,7 +7,7 @@ def initialize(state)
     attr_reader :state
 
     def pending?
-      state == :pending
+      state == :pending || state == ::RDF::URI("http://fedora.info/definitions/1/0/access/ObjState#inactive")
```
jcoyne: I don't think we need this earlier access at all. In fact, I'm not sure we need this class at all.  Can we just move this check into the indexer?  Also use `Vocab::FedoraResourceStatus.inactive` instead of creating a new URI.

app/search_builders/curation_concerns/filter_by_type.rb
```diff
@@ -3,7 +3,7 @@ module FilterByType
     extend ActiveSupport::Concern
 
     included do
-      self.default_processor_chain += [:filter_models]
+      self.default_processor_chain += [:filter_models, :only_active_works]
```
jcoyne: I don't think this belongs in `FilterByType`  Instead can you make a  new module called `FilterByStatus`?

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -198,6 +198,10 @@ def curation_concern_from_search_results
     def search_result_document(search_params)
       _, document_list = search_results(search_params)
       raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
+      if document_list.present? && document_list.first["suppressed_bsi"]
```
jcoyne: What about the users who are supposed to take action on this work?

app/search_builders/curation_concerns/filter_by_type.rb
luisgreg99: @jcoyne: @straleyb is addressing this change in his WIP PR https://github.com/projecthydra/curation_concerns/pull/1085 He is calling the module `FilterSuppressed`

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -198,6 +198,10 @@ def curation_concern_from_search_results
     def search_result_document(search_params)
       _, document_list = search_results(search_params)
       raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
+      if current_user.blank? && document_list.present? && document_list.first["suppressed_bsi"]
```
jcoyne: Why would the `current_user` be `blank?` why not just test for `current_user`? Can you please combine these three checks into a new method like `document_suppressed?(document_list)`

luisgreg99: I was using `blank?` to display the message only for anonymous users, should I use `unless` instead of blank?
```
unless current_user
  document_list.present? && document_list.first["suppressed_bsi"]
end
```

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -194,7 +194,17 @@ def curation_concern_from_search_results
...
         raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
+        if document_suppressed?(document_list)
+          flash[:notice] = I18n.t("curation_concerns.workflow.default.unauthorized")
+          raise CanCan::AccessDenied.new(nil, :show)
+        end
         document_list.first
       end
+
+      def document_suppressed?(document_list)
+        unless current_user
```
jcoyne: This is not an anonymous user. If you are looking for anonymous users, you probably want to check for `guest?` Keep in mind that guest users exist and have `guest?` only if the `devise-guest` gem is being used, otherwise, you'd check for anonymous by checking `current_user.new_record?`

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -194,7 +194,17 @@ def curation_concern_from_search_results
...
         _, document_list = search_results(search_params)
         raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
+        if document_suppressed?(document_list)
+          flash[:notice] = I18n.t("curation_concerns.workflow.default.unauthorized")
+          raise CanCan::AccessDenied.new(nil, :show)
+        end
         document_list.first
       end
+
+      def document_suppressed?(document_list)
```
jcoyne: Add method documentation.

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
jcoyne: I'm wondering why we're only suppressing for anonymous users. Shouldn't we suppress for all users who do not have a current workflow role on the document?

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -201,7 +201,19 @@ def curation_concern_from_search_results
...
+        document = document_list.first
+        if user_cannot_access_due_to_workflow?(document)
+          flash[:notice] = I18n.t("curation_concerns.workflow.default.unauthorized")
+          render 'unauthorized', status: :unauthorized
+        end
+        document
+      end
+
+      # If document is currently suppressed in workflow, only display
+      # the document if the user has a role in the workflow
```
jcoyne: This should probably be `if the user can perform an action for the current workflow state`

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -201,7 +201,19 @@ def curation_concern_from_search_results
       def search_result_document(search_params)
         _, document_list = search_results(search_params)
         raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
-        document_list.first
+        document = document_list.first
+        if user_cannot_access_due_to_workflow?(document)
+          flash[:notice] = I18n.t("curation_concerns.workflow.default.unauthorized")
```
jcoyne: It's using `default` workflow here, even though this may not be the default workflow. If we aren't going to use the workflow name, I'd go with `curation_concerns.workflow.unauthorized`

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -201,7 +201,19 @@ def curation_concern_from_search_results
...
+          render 'unauthorized', status: :unauthorized
```
jcoyne: The unauthorized template is going to display a message about it being private, which I though we were trying to avoid here: https://github.com/projecthydra/curation_concerns/blob/master/app/views/curation_concerns/base/unauthorized.html.erb

mjgiarlo:
@jcoyne I addressed your requested changes and rejiggered the impl a bit. I'm running into two failures that I don't understand; if you get a couple minutes, would you mind taking a look at the Travis failures?

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -203,5 +204,22 @@ def search_result_document(search_params)
         raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
         document_list.first
       end
+
+      def authorize_workflow_role!
+        concern = _curation_concern_type.find(params[:id])
```
jcoyne: This is hitting Fedora, which is slow.  Do a query against solr instead.

mjgiarlo:
@jcoyne I believe the changes I've made address your review.

Note though that this pulls in the `feature/SuppressObjectForWorkflow` branch which adds the ability to ask Solr docs if they are suppressed. I've pushed this branch up to show that the approach works, but before it can be merged, I will need to disentangle the two branches once https://github.com/projecthydra/curation_concerns/pull/1103 is merged, and then rebase on master

In the meantime, @jcoyne, are you inclined to re-review this PR?

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -203,5 +204,20 @@ def search_result_document(search_params)
         raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
         document_list.first
       end
+
+      def authorize_workflow_role!
+        if user_cannot_access_due_to_workflow?(presenter.solr_document)
```
jcoyne: Will this presenter ever find anything? Won't the search builder filter out the suppressed results?

mjgiarlo: Does the work show view call a search builder?

mjgiarlo: It doesn't look like the search builder is invoked here.

mjgiarlo:
@jcoyne :green_heart: This is ready for another review.

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
jcoyne: `presenter` calls `curation_concern_from_search_results` which calls `search_result_document` which calls `search_results` which calls `search_builder_class`

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
```diff
@@ -10,8 +10,9 @@ module FilterSuppressedWithRoles
     include CurationConcerns::FilterSuppressed
 
     def only_active_works(solr_parameters)
-      # Do not filter works if current user is permitted to take workflow actions
-      return if user_has_active_workflow_role?
+      # Skip the filter if work is not suppressed or if the current user
+      # is permitted to take workflow actions
+      return if work_not_suppressed? || user_has_active_workflow_role?
```
jcoyne: is the `work_not_suppressed?` check required here?  Won't the super just take care of this?

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -203,5 +204,20 @@ def search_result_document(search_params)
...
+          flash[:notice] = I18n.t("curation_concerns.workflow.unauthorized")
+          render 'unavailable', status: :unauthorized
```
jcoyne: What if they want `json` or some other non-html type?

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -77,6 +78,64 @@
       end
     end
 
+    context 'with unsuppressed work' do
+      let(:work) { create(:public_generic_work) }
+      let(:document) { SolrDocument.new(work.to_solr) }
+      let(:presenter) { CurationConcerns::WorkShowPresenter.new(document, nil) }
+      before do
+        allow_any_instance_of(GenericWork).to receive(:suppressed?).and_return(false)
```
jcoyne: Why is there an instance of a GenericWork here? Shouldn't we just be dealing in the realm of Solr?

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -77,6 +78,64 @@
...
+        allow(controller).to receive(:presenter).and_return(presenter)
+      end
+      it 'renders the show view and does not use active_works filter' do
+        expect(controller).not_to receive(:user_has_active_workflow_role?)
```
jcoyne: This method is on the search builder, not the controller.

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -77,6 +78,64 @@
...
+    end
+
+    context 'with work still in workflow' do
+      let(:workflow_user) { create(:user) }
+      let(:work) { create(:public_generic_work) }
+      let(:permission_scope) { [] }
+      let(:document) { SolrDocument.new(work.to_solr) }
+      let(:presenter) { CurationConcerns::WorkShowPresenter.new(document, nil) }
+      before do
+        sign_out user
```
jcoyne: This doesn't do anything on a controller test, does it?

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -77,6 +78,64 @@
...
+        allow_any_instance_of(GenericWork).to receive(:suppressed?).and_return(true)
```
jcoyne: We should avoid instances of AF::Base in `#show`

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -203,5 +204,20 @@ def search_result_document(search_params)
         raise CanCan::AccessDenied.new(nil, :show) if document_list.empty?
         document_list.first
       end
+
+      def authorize_workflow_role!
```
jcoyne: It seems to me that we should have `user_cannot_access_due_to_workflow?` in Ability (or at least not the controller).  Since CanCan is doing authorization with exceptions, should we be following that pattern with `authorize_workflow_role!` (and do `rescue_from WorkflowException, with: :my_exception handler?`

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -77,6 +78,64 @@
...
+      context 'with an anonymous user' do
+        let(:workflow_user) { nil }
+        it 'shows the unauthorized message' do
+          get :show, params: { id: work }
+          expect(response.code).to eq '401'
+          expect(response).to render_template(:unavailable)
+          expect(flash[:notice]).to eq 'The work is not currently available because it has not yet completed the approval process'
+        end
+      end
+      context 'with a user lacking workflow permission' do
```
jcoyne: How is this test any different than the one above?

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -56,10 +56,11 @@
         let(:presenter) { double }
         before do
           allow(controller).to receive(:presenter).and_return(presenter)
+          allow(controller).to receive(:authorize_workflow_role!).and_return(nil)
```
jcoyne: Why are we doing this here?  Should this be reflected in a change to the context?

app/controllers/concerns/curation_concerns/curation_concern_controller.rb
mjgiarlo: Doesn't CanCan raise `CanCan::AccessDenied`? Not sure how to get CanCan to raise `WorkflowException`.

mjgiarlo:
@jcoyne I made some/many of the changes you requested, but this work was rushed and tbh I'm not sure I'm on the right track. So if things are broken or seem half-done, it's only because they are. Could use some pointers tomorrow. Thx for the review.

mjgiarlo:
@jcoyne re-review? :)

PR #1081, opened 2016-11-05
Start workflow with an action

jcoyne:
This allows a deposit to run an action or a notification

### DIFF
--- a/app/services/curation_concerns/actors/actor_factory.rb
+++ b/app/services/curation_concerns/actors/actor_factory.rb
@@ -15,8 +15,8 @@ module CurationConcerns
          ApplyOrderActor,
          InterpretVisibilityActor,
          GrantEditToDepositorActor,
-         InitializeWorkflowActor,
-         model_actor(curation_concern)]
+         model_actor(curation_concern),
+         InitializeWorkflowActor]
       end
 
       def self.model_actor(curation_concern)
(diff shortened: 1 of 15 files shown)


app/services/curation_concerns/actors/actor_factory.rb
```diff
@@ -15,8 +15,8 @@ def self.stack_actors(curation_concern)
          ApplyOrderActor,
          InterpretVisibilityActor,
          GrantEditToDepositorActor,
-         InitializeWorkflowActor,
-         model_actor(curation_concern)]
+         model_actor(curation_concern),
+         InitializeWorkflowActor]
```
straleyb: I see you changed the order of these actors as well. Im curious as to what was going on here with the order of actions and what that might have been causing.

jcoyne: We need an identifier to be minted before we can initialize workflow.

app/services/curation_concerns/workflow/workflow_action_service.rb
```diff
@@ -0,0 +1,56 @@
+module CurationConcerns
+  module Workflow
+    class WorkflowActionService
```
straleyb: Did we want a test for this?

straleyb:
Just had a few questions. Otherwise its good for me. Do we need final approval from others before merging this? @mjgiarlo @jcoyne

PR #1082, opened 2016-11-08
url_for_document should return a routable entity, not a string

jcoyne:
So that the value of url_for_document can be passed to polymorphic_url
as seen here:
https://github.com/projectblacklight/blacklight/blob/e7d7cb883e4d24a5263797cac47a8e5fd5b603ca/app/views/catalog/_document.rss.builder#L3


cbeer:
Hm. That also seems like a blacklight bug. `polymorphic_url` was probably the wrong choice?

jcoyne:
@cbeer I don't know, that seems exactly the correct usage according to http://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html.  Or are you saying that `url_for_document` should be returning a string?

mjgiarlo:
In the meantime, @cbeer, are you cool with this PR being merged? Or do you think we should be reaching down into Blacklight here?

PR #1084, opened 2016-11-08
Allow groups to be cast to Sipity::Agents

jcoyne:
I did it all for @tpendragon 😉

### DIFF
--- a/app/models/curation_concerns/group.rb
+++ b/app/models/curation_concerns/group.rb
@@ -1,0 +1,23 @@
+module CurationConcerns
+  class Group
+    def initialize(name)
+      @name = name
+    end
+
+    attr_reader :name
+
+    def to_sipity_agent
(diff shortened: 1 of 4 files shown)


app/models/curation_concerns/group.rb
```diff
@@ -0,0 +1,23 @@
+module CurationConcerns
+  class Group
+    def initialize(name)
+      @name = name
+    end
+
+    attr_reader :name
+
+    def to_sipity_agent
+      sipity_agent || create_sipity_agent!
```
mjgiarlo: Why not just

```ruby
Sipity::Agent.find_or_create_by!(proxy_for_id: name, proxy_for_type: self.class.name)
```

?

jcoyne: I could have done that. I was just following the patter @jeremyf used for `User`s.

PR #1085, opened 2016-11-08
[WIP] Feature/hide suppressed

straleyb:
This is a work in progress. This gets items that are suppressed hidden from search results. This does however cause works to not be viewable by anyone. As an admin, one should be able to see these items and push them through the workflow.

### DIFF
--- a/app/models/concerns/curation_concerns/publishable.rb
+++ b/app/models/concerns/curation_concerns/publishable.rb
@@ -14,7 +14,13 @@ module CurationConcerns
     # Override this method if you have some critera by which records should not
     # display in the search results.
     def suppressed?
+      state ||= to_workflow_name
       state_workflow.new(state).pending?
+    end
+
+    def to_workflow_name
+      raise "Can't create an entity until the model has been persisted" unless persisted?
+      @sipity_workflow_name ||= Sipity::Entity.find_by(proxy_for_global_id: to_global_id.to_s).workflow_state_name unless Sipity::Entity.find_by(proxy_for_global_id: to_global_id.to_s).nil?
     end
 
     def to_sipity_entity
--- a/app/models/curation_concerns/state_workflow.rb
+++ b/app/models/curation_concerns/state_workflow.rb
@@ -7,7 +7,7 @@ module CurationConcerns
     attr_reader :state
 
     def pending?
-      state == :pending
+      state == :pending || state == ::RDF::URI("http://fedora.info/definitions/1/0/access/ObjState#inactive")
     end
   end
 end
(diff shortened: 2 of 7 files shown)


app/models/concerns/curation_concerns/publishable.rb
```diff
@@ -14,9 +14,15 @@ module Publishable
     # Override this method if you have some critera by which records should not
     # display in the search results.
     def suppressed?
+      state ||= to_workflow_name
       state_workflow.new(state).pending?
     end
 
+    def to_workflow_name
+      raise "Can't create an entity until the model has been persisted" unless persisted?
+      @sipity_workflow_name ||= Sipity::Entity.find_by(proxy_for_global_id: to_global_id.to_s).workflow_state_name unless Sipity::Entity.find_by(proxy_for_global_id: to_global_id.to_s).nil?
```
jcoyne: This causes 2 find_by calls.  Can you rewrite to do the find only once?

app/models/curation_concerns/state_workflow.rb
```diff
@@ -7,7 +7,7 @@ def initialize(state)
     attr_reader :state
 
     def pending?
-      state == :pending
+      state == :pending || state == ::RDF::URI("http://fedora.info/definitions/1/0/access/ObjState#inactive")
```
jcoyne: Use `Vocab::FedoraResourceStatus.inactive` instead of initializing a new URI.   Is the `state == :pending` is still relevant after this change?

app/search_builders/curation_concerns/collection_search_builder.rb
```diff
@@ -1,6 +1,7 @@
 module CurationConcerns
   class CollectionSearchBuilder < ::SearchBuilder
     include FilterByType
+    include FilterSuppressed
```
jcoyne: Would a collection ever be suppressed?

app/search_builders/curation_concerns/filter_suppressed.rb
```diff
@@ -0,0 +1,14 @@
...
+  module FilterSuppressed
+    extend ActiveSupport::Concern
+
+    included do
+      self.default_processor_chain += [:only_active_works]
+    end
+
+    def only_active_works(solr_parameters)
+      solr_parameters[:fq] ||= []
+      solr_parameters[:fq] << '-suppressed_bsi:true'
```
jcoyne: Wouldn't we only want this filter only for certain users?

straleyb:
@jcoyne Im gonna close this PR. Now that we have the state being set even on the pending state, a lot of this work can be changed or removed. Im gonna create a new branch and PR with my work thats being done on filtering suppressed works depending on the workflow role the user has.

PR #1097, opened 2016-11-16
reload work to get updates that could have been made by workflow acti…

### LINKED ISSUE #1096: [bug] WorkflowActionService can cause sipity entity and work to become out of sync
A WorkflowAction, such as ActivateObject or DeactivateObject cause cause an update to the sipity entity (changing state: https://github.com/projecthydra/curation_concerns/blob/1be404f895c71292ed2614d26022c36b964a9b3b/app/services/curation_concerns/workflow/activate_object.rb#L5) which persists to SOLR. 

`subject.work` becomes out of sync and has previously set state causing SOLR to change during the `update_index` call at: https://github.com/projecthydra/curation_concerns/blob/1be404f895c71292ed2614d26022c36b964a9b3b/app/services/curation_concerns/workflow/workflow_action_service.rb#L21

`subject.work.reload` refreshes with changes that could have been made by the WorkflowAction

revgum:
Fixes #1096

### DIFF
--- a/app/services/curation_concerns/workflow/workflow_action_service.rb
+++ b/app/services/curation_concerns/workflow/workflow_action_service.rb
@@ -18,6 +18,7 @@ module CurationConcerns
         comment = create_sipity_comment
         handle_sipity_notifications(comment: comment)
         handle_additional_sipity_workflow_action_processing(comment: comment)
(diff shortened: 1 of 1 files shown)


jcoyne:
Can you write a test that shows why this change is required?

PR #1099, opened 2016-11-17
Add a built in function to grant edit access to the depositor

### LINKED ISSUE #1095: Zero-step (default) workflow should grant edit access to the depositor
In version 1.6.0 of CC, the depositor is granted edit access.  Thus we must change the default workflow to grant edit access to the depositing user.

### LINKED ISSUE #1096: [bug] WorkflowActionService can cause sipity entity and work to become out of sync

jcoyne:
Fixes #1095
This also raised the issue where the callback methods were not indexing
the correct instance of the work.
Fixes #1096
Supersedes #1097


jcoyne:
Backported to 1-7-stable as fb2d8ce9f6ad41ced74fb86efcefeeabf98374ac

PR #1103, opened 2016-11-22
Add workflow-role based authorization

straleyb:
Add the ability to allow a user to see or not see a work based on their workflow roles and the current workflow state.


jcoyne:
I don't understand the purpose of adding entities to these feature tests. This just slows down the test suite.

straleyb:
@jcoyne Sipity based access control relies on a Sipity::Entity existing in the database for a workflow. This is the line of code here https://github.com/projecthydra/curation_concerns/pull/1103/files#diff-aede147ec36a958b2d6485797b8b15c6R19

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
```diff
@@ -0,0 +1,27 @@
+module CurationConcerns
+  module FilterSuppressedWithRoles
+    extend CurationConcerns::FilterSuppressed
```
jcoyne: I don't think this is doing what you think it's doing

app/models/concerns/curation_concerns/suppressible.rb
```diff
@@ -1,20 +1,25 @@
...
+    end
+
+    def deactivate
+      self.state = Vocab::FedoraResourceStatus.inactive
+    end
 
-      class_attribute :state_workflow, instance_writer: false
-      self.state_workflow = StateWorkflow
+    def activate
+      self.state = Vocab::FedoraResourceStatus.active
```
jcoyne: I think this should remain in the ActivateObject service class. Why are you moving this here. The responsibity of the model is persisting data. The responsibility of the service object is knowing how to activate. Each object should have a single responsibility.

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -4,7 +4,16 @@
 # which is included into .internal_test_app/app/controllers/generic_works_controller.rb
 describe CurationConcerns::GenericWorksController do
   let(:user) { create(:user) }
-  before { sign_in user }
+  let(:work) { create(:generic_work, user: user) }
+  let(:sipity_entity) do
```
mjgiarlo: If you change this `let` to `let!`, you can remove `sipity_entity` on line 15 below.

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -4,7 +4,16 @@
...
+    create(:sipity_entity,
+           proxy_for_global_id: work.to_global_id.to_s,
+           workflow_state_id: 2)
```
mjgiarlo: As discussed in Slack, you should be able to :fire: all the hard-coded `workflow_state_id` stuff you added.

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -4,7 +4,16 @@
...
+  end
+  before do
+    sign_in user
```
mjgiarlo: Why would this be required now, if it wasn't required before?

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -16,7 +25,13 @@
 
       context "with a parent work" do
         let(:parent) { create(:generic_work, title: ['Parent Work'], user: user, ordered_members: [work]) }
+        let(:parent_sipity_entity) do
```
mjgiarlo: If you use `let!` instead of `let`, you can dump the change to line 34 below.

spec/controllers/curation_concerns/generic_works_controller_spec.rb
```diff
@@ -238,55 +252,56 @@
   end
 
   describe '#destroy' do
-    let(:work_to_be_deleted) { create(:private_generic_work, user: user) }
```
mjgiarlo: Why rename the variable here? Do you want this `work` to stand in for the `work` created on line 7?

spec/factories/generic_works.rb
```diff
@@ -4,6 +4,7 @@
   factory :generic_work, aliases: [:work, :private_generic_work], class: GenericWork do
     transient do
       user { FactoryGirl.create(:user) }
+      # workflow_state { FactoryGirl.create(:workflow_state) }
```
mjgiarlo: Remove if not needed when you're ready to submit the PR.

spec/factories/generic_works.rb
```diff
@@ -13,6 +14,10 @@
       work.apply_depositor_metadata(evaluator.user.user_key)
     end
 
+    # after(:create) do |work, evaluator|
```
mjgiarlo: Remove these lines if not needed when you're ready to submit the PR.

spec/factories/workflows.rb
```diff
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :workflow, class: Sipity::Workflow do
-    name 'generic_work'
+    sequence(:name) { |n| "generic_work-#{n}" }
+    # name 'generic_work'
```
mjgiarlo: Remove if not needed when you're ready to submit the PR.

spec/factories/workflows.rb
```diff
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :workflow, class: Sipity::Workflow do
-    name 'generic_work'
+    sequence(:name) { |n| "generic_work-#{n}" }
```
mjgiarlo: Is this change still needed?

spec/features/add_file_spec.rb
```diff
@@ -3,9 +3,15 @@
 feature 'Add an attached file' do
   let(:user) { create(:user) }
   let!(:work) { create(:work, user: user) }
+  let(:sipity_entity) do
```
mjgiarlo: Use `let!` instead of `let` and you can remove line 14 below.

spec/features/add_file_spec.rb
```diff
@@ -3,9 +3,15 @@
...
+    create(:sipity_entity,
+           proxy_for_global_id: work.to_global_id.to_s,
+           workflow_state_id: 2)
```
mjgiarlo: :fire:

spec/features/create_child_work_spec.rb
```diff
@@ -3,7 +3,11 @@
 
 feature 'Creating a new child Work', :workflow do
   let(:user) { FactoryGirl.create(:user) }
-
+  let(:sipity_entity) do
+    create(:sipity_entity,
```
mjgiarlo: Same re: above: `let!`

spec/features/create_child_work_spec.rb
```diff
@@ -3,7 +3,11 @@
...
+           proxy_for_global_id: parent.to_global_id.to_s,
+           workflow_state_id: 2)
```
mjgiarlo: :fire:

spec/models/generic_work_spec.rb
```diff
@@ -37,17 +37,36 @@
...
       it { is_expected.to be false }
     end
   end
 
+  context "when activating or deactivating a work" do
+    let(:work) { described_class.new }
+    describe "#activate" do
+      subject { work.activate }
+      it "sets its state to active" do
+        subject
```
mjgiarlo: Not sure what line 56 is adding. Can you rework this test to be more succinct? Like

```ruby
describe '#activate' do
  subject { work.activate }
  it { is_expected.not_to be_suppressed }
end
```

spec/models/generic_work_spec.rb
```diff
@@ -37,17 +37,36 @@
...
+        expect(work.suppressed?).to be false
+      end
+    end
+
+    describe "#deactivate" do
+      subject { work.deactivate }
+      it "sets its state to inactive" do
+        subject
```
mjgiarlo: Nuke line 64 and rework as above:

```ruby
describe '#deactivate' do
  subject { work.deactivate }
  it { is_expected.to be_suppressed }
end
```

spec/services/curation_concerns/workflow/permission_query_spec.rb
```diff
@@ -24,7 +24,7 @@ module Workflow
                                workflow_state: PowerConverter.convert_to_sipity_workflow_state('initial', scope: sipity_workflow)
                               )
       end
-      let(:sipity_workflow) { Sipity::Workflow.first }
+      let(:sipity_workflow) { Sipity::Workflow.where(name: 'testing').first }
```
mjgiarlo: Is this still needed? Are there > 1 workflows?

spec/services/curation_concerns/workflow/status_list_service_spec.rb
```diff
@@ -4,6 +4,7 @@
   describe "#each" do
     let(:user) { create(:user) }
     let(:service) { described_class.new(user) }
+    let(:sipity_entity) { instance_double(Sipity::Entity, workflow: workflow, workflow_state: workflow_state) }
```
mjgiarlo: Why is this test using instance doubles when the others are using factories? If you can make this test use the same factories, that would bring extra consistency.

spec/services/curation_concerns/workflow/status_list_service_spec.rb
```diff
@@ -14,16 +15,21 @@
...
     before do
-      create(:sipity_entity)
       ActiveFedora::SolrService.add([document, ability], commit: true)
     end
 
     let(:results) { service.each.to_a }
 
     context "when user has roles" do
       before do
+        sipity_entity
```
mjgiarlo: Use `let!` above instead.

spec/controllers/curation_concerns/generic_works_controller_spec.rb
jcoyne: It's better to use a before block though. Otherwise rubocop rspec will complain.

mjgiarlo: @jcoyne is that so? There's lots of `let!` in the current codebase.

mjgiarlo: @jcoyne :ok_hand:

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
straleyb: So my idea of this was, creating an object that inherits the methods from `CurationConcerns::FilterSuppressed` and overrides them to cut down on code duplication. Would you just include the module instead of extend the module?

straleyb:
@jcoyne Based on your comment about creating objects slowing down the test suite, I hadn't noticed a slow down personally, but would you much rather mocking the entities instead? I had the assumption they needed to exist in the database for the query here. https://github.com/projecthydra/curation_concerns/pull/1103/files#diff-aede147ec36a958b2d6485797b8b15c6R20

spec/controllers/curation_concerns/generic_works_controller_spec.rb
straleyb: So I changed the name of this variable for consistency sake. Thats just that many more places you can miss a variable and confused me several times. So I figured it would be nice to go with a more consistent name. But I can change it back no biggy. Was just a proposition really.

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def only_active_works(solr_parameters)
+      if current_work.respond_to?(:to_sipity_entity) # Used because Collections use WorkSearchBuilder to filter down to a single item
+        super if user_work_permissions.empty?
```
mjgiarlo: Per @jcoyne's above comment, the `super`s in this module do not wind up calling the method with the same name in `FilterSuppressed` unless the class you're calling this method on *also* includes `FilterSuppressed`.

spec/factories/workflows.rb
straleyb: Yes, this change is needed. This is because now that we are creating sipity entities, these workflows need to exist and not having this will throw the `UNIQUE constraint failed` errors in the DB

jcoyne:
@straleyb Okay, I wasn't sure why that was required. I think you could test out just the SearchBuilder itself with real objects, and then mock out the response from the search builder on the Controller tests.  Otherwise add comments to the test about why those are needed.

spec/controllers/curation_concerns/generic_works_controller_spec.rb
mjgiarlo: I'm OK with this change.

straleyb:
@jcoyne Sounds good. Thanks for the clarification. I can go through and rework those.

spec/factories/workflows.rb
mjgiarlo: :ok_hand: sounds good

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
straleyb: So would the change here be to `include` FilterSuppressed rather than extend it? Or is there another pattern or technique I should be using here?

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
mjgiarlo: @straleyb I believe that should do the trick.

mjgiarlo:
@jcoyne when you get a minute, could you re-review this PR? It's :green_heart:

app/models/concerns/curation_concerns/suppressible.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    ##
+    # Used to restrict visibility on search results for a work that is inactive. If the state is not set, the
+    # default behavior is to consider the work not to be suppressed.
+    #
+    # Override this method if you have some criteria by which records should not display in the search results.
+    def suppressed?
+      return false if state.nil?
+      # This is a clumsy check only needed under RDF < 2.0
+      if ENV['RDF_VERSION'] < '2.0'
```
jcoyne: This will only be set on Travis. It won't be suitable in production

mjgiarlo: Dang. Shall I ask RDF.rb its version instead?

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
```diff
@@ -0,0 +1,28 @@
...
+      else
+        super
+      end
+    end
+
+    private
+
+      def current_work
+        ActiveFedora::Base.find(blacklight_params[:id])
```
jcoyne: Loading from Fedora is a slow operation, which SearchBuilders were created to avoid.  SearchBuilders should never interface with Fedora.

app/models/concerns/curation_concerns/suppressible.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    ##
+    # Used to restrict visibility on search results for a work that is inactive. If the state is not set, the
+    # default behavior is to consider the work not to be suppressed.
+    #
+    # Override this method if you have some criteria by which records should not display in the search results.
+    def suppressed?
+      return false if state.nil?
+      # This is a clumsy check only needed under RDF < 2.0 where URIs
+      # are returned as AT::Resources
```
jcoyne: I believe both versions return `AT::Resources`, but the latter versions implement `==` correctly

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
```diff
@@ -0,0 +1,28 @@
+module CurationConcerns
+  module FilterSuppressedWithRoles
```
jcoyne: Add a module comment about when this can be used and its assumptions (`solr_params[:id]` being one).

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
```diff
@@ -0,0 +1,28 @@
...
+    extend ActiveSupport::Concern
+    include CurationConcerns::FilterSuppressed
+
+    def only_active_works(solr_parameters)
+      if current_work.respond_to?(:to_sipity_entity) # Used because Collections use WorkSearchBuilder to filter down to a single item
```
jcoyne: Maybe the Collections shouldn't use WorkSearchBuilder. That's probably better than adding these conditionals.

app/services/curation_concerns/workflow/activate_object.rb
```diff
@@ -1,5 +1,9 @@
 module CurationConcerns
   module Workflow
+    # This is a built in function for workflow, so that a workflow action can be created that
+    # activates an object.
+    # Target is an instance of a model with CurationConcerns::Suppressible
```
jcoyne: This should be YARD doc on the `.call` method.

app/services/curation_concerns/workflow/deactivate_object.rb
```diff
@@ -2,6 +2,8 @@ module CurationConcerns
   module Workflow
     # This is a built in function for workflow, so that a workflow action can be created that
     # deactivates an object.
+    # Target is an instance of a model with CurationConcerns::Suppressible
```
jcoyne: This should be YARD doc on the `.call` method.

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
jcoyne: Actually, I'm not seeing where this is happening.  Current_work should always be an instance of a work. Can we remove this conditional?

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
mjgiarlo: Done.

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
mjgiarlo: Removed.

app/services/curation_concerns/workflow/activate_object.rb
mjgiarlo: Done.

app/services/curation_concerns/workflow/deactivate_object.rb
mjgiarlo: Done.

app/search_builders/curation_concerns/filter_suppressed_with_roles.rb
```diff
@@ -0,0 +1,26 @@
...
+  # actions for the work's current state
+  #
+  # Assumes presence of `blacklight_params[:id]` and a SolrDocument
+  # corresponding to that `:id` value
+  module FilterSuppressedWithRoles
+    extend ActiveSupport::Concern
+    include CurationConcerns::FilterSuppressed
+
+    def only_active_works(solr_parameters)
+      super if user_work_permissions.empty?
```
jcoyne: Can we put this in a method called `current_user_has_active_role?` and then do: `return super unless current_user_has_active_role?`

mjgiarlo: Will do.

PR #1107, opened 2016-11-23
Changing derivatives job to perform_now

cam156:
This allows for characterization and derivatives generation to be performed in the same context,
which is safer if a file is being passed in.

To give some context for why I am asking for this change I am calling a perform_now on characterization passing in a file name, and the deleting the file that I am passing to characterization.  The problem is the perform_later for derivatives then does not get access to the file.

### DIFF
--- a/app/jobs/characterize_job.rb
+++ b/app/jobs/characterize_job.rb
@@ -12,6 +12,6 @@ class CharacterizeJob < ActiveJob::Base
     file_set.characterization_proxy.save!
     file_set.update_index
     file_set.parent.in_collections.each(&:update_index) if file_set.parent
-    CreateDerivativesJob.perform_later(file_set, file_id, filename)
+    CreateDerivativesJob.perform_now(file_set, file_id, filename)
   end
 end
--- a/spec/jobs/characterize_job_spec.rb
+++ b/spec/jobs/characterize_job_spec.rb
@@ -25,7 +25,7 @@ describe CharacterizeJob do
       expect(Hydra::Works::CharacterizationService).to receive(:run).with(file, filename)
       expect(file).to receive(:save!)
       expect(file_set).to receive(:update_index)
-      expect(CreateDerivativesJob).to receive(:perform_later).with(file_set, file.id, filename)
+      expect(CreateDerivativesJob).to receive(:perform_now).with(file_set, file.id, filename)
       described_class.perform_now(file_set, file.id)
     end
   end
@@ -46,7 +46,7 @@ describe CharacterizeJob do
       allow(Hydra::Works::CharacterizationService).to receive(:run).with(file, filename)
       allow(file).to receive(:save!)
       allow(file_set).to receive(:update_index)
-      allow(CreateDerivativesJob).to receive(:perform_later).with(file_set, file.id, filename)
+      allow(CreateDerivativesJob).to receive(:perform_now).with(file_set, file.id, filename)
     end
     it "reindexes the collection" do
       expect(collection).to receive(:update_index)


jcoyne:
@cam156 where are you deleting the file? How are you ensuring you the file is not deleted before characterization takes place?

cam156:
@jcoyne this is during the migration process, so we are downloading the file from the old Fedora, and then performing the characterization job inline.

jcoyne:
If the characterization job is run by the inline adapter, then surely the derivatives job is run by the inline adapter too, right?

cam156:
@jcoyne We are calling perform_now to run it inline, not using a different adapter.  Calling perform_now on characterization still allows perform_later on derivatives generation.

jcoyne:
My concern is that now there is no way to isolate derivatives from characterization.  I'm imagining the case where we want to run the derivatives in a different queue or worker machine

cam156:
@jcoyne with the file option that is not really safe.  That said I can see that as a possible use case.  Does anyone currently have that use case?  I know people have the use case of wanting to use an existing file beyond us.

escowles:
We use a network-mounted temp space to get around this issue — the ingest, characterization and derivatives jobs may run on separate machines, but they can all use the local file on the shared temp space so they don't have to re-download the file from Fedora.  But I wonder if there is any way to have a chain of jobs run on the same worker, without having to run them all inline?

The other way to handle this would be to have a configuration option and/or place to override whether the jobs are run with `perform_now` or `perform_later`.  I can imagine different applications having different intentions about, e.g., whether to run all the ingest jobs first, or ingest a file and immediately run the characterization and derivatives before ingesting other files, etc.

cam156:
@escowles how do you clean up the file after the job completes?  That is part of the reason we running now vs later.  Having a configuration option would be a good middle ground.  
 
I also have often wondered how to reliably chain separate jobs other than just performing them within the job.  

I wonder if it should really be two separate jobs, one job that loads the file from fedora and runs the characterization and derivatives, and one job that loads the file from the file system.

escowles:
@cam156 Right now, we just have a cron job that deletes files older than 24 hours to cleanup old files.  We've talked about updating our derivatives job to delete the file, but haven't done that.

jcoyne:
How about having another event fire when the derivative is complete?  Then you could attach something to that event to remove the file.

jcoyne:
@escowles we already have that configuration option, you just set the runner to the inline adapter:

```ruby
Rails.application.config.active_job.queue_adapter = :inline
```
http://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/InlineAdapter.html

escowles:
@jcoyne But the InlineAdapter runs the jobs immediately within the webapp, right?  I mean that you want the jobs to be run asynchronously, but only child jobs to be inline.

jcoyne:
@escowles I believe @cam156 is doing this in a batch job. She made this change, such that she could immediately delete the original file. In order to know that she can safely delete the file, she has to know when the characterization is done.  As far as I know, the only way she could do that is by running the jobs inline.

awead:
I've been wondering if would should not chain the jobs, and have the IngestFileJob call perform_later on both characterization and derivatives. That way, they can both be run in parallel and speed things up a bit.

It also frees up things so folks can run either individually, without having to run the other. In @cam156 case, she could pass in the file to both jobs, calling perform_now, and then delete the file afterwards. We could also create a new job, CharacterizeAndCreateDerivatives, that simply repackages the lines from IngestFileJob that call those jobs.

escowles:
@awead I thought that the derivative job required the characterization job to be run first, to get the mime type, etc. to drive the derivative logic.

awead:
@escowles ah, you might be right there, but I don't know how critical that is because a mime type is set here in the ingest job:

https://github.com/projecthydra/curation_concerns/blob/master/app/jobs/ingest_file_job.rb#L15

but it could be nil. I'm not under what circumstances there would not be a mime type.

jcoyne:
@awead the ingest job is setting a mime type based on what the browser sent. It could be completely useless or wrong.  The characterization is going to to a better job at setting the true mime type.

awead:
So I'm seeing basic use cases: 1) ingesting files, and 2) piecemeal. Since characterization and derivatives seems to be tied together because of the mime type issue, I'm thinking there ought to be one unified CharacterizeAndCreateDerivatives job that the ingest calls for new files, and then individual jobs for just characterize and derivatives. The unified job could simply call the other two using perform_now. I think that would greatly clear up the confusion and hopefully offer easier solutions for the piecemeal use cases.

hackmastera:
@awead @jcoyne @cam156 @escowles It doesn't look like an issue ever came out of this discussion. 

Use cases are:
- batch ingest where files should be deleted once they're no longer in use
- UI-based ingest where jobs may run on different queues and/or different machines
- UI-based ingest where files are deleted as they go so you don't have to run a cleanup cron

The options above that weren't vetoed are:

- Add an event to the derivatives job such that the file could be deleted on its completion
- Add a configuration option for perform_now or perform_later
- Create three jobs, to allow more flexibility: CharacterizeAndCreateDerivatives (one calls the other via perform_now), Characterize (doesn't call CreateDerivatives?), and CreateDerivatives (already exists). I'm not actually sure this accommodates the use cases, though.

There's another option of doing nothing and saying we don't really support batch loading via scripts in this way. But I think a lot of people actually want to load content this way so it might be nice to be able to accommodate it.

Thoughts? Anything else I missed?

PR #1105, opened 2016-11-23
Collection search builder

atz:
Salvaging work from #968

PR #1110, opened 2016-11-29
Collection search builder part2

atz:
Retargeted PR against master after part1 #1105 merged.


app/controllers/concerns/curation_concerns/selects_collections.rb
```diff
@@ -61,8 +61,9 @@ def collections_search_builder_class
   end
 
   def collections_search_builder(access_level = nil)
-    collections_search_builder_class.new(self).tap do |builder|
-      builder.discovery_perms = access_levels[access_level] if access_level
-    end
+    collections_search_builder_class.new(self, access_levels[access_level])
+    # collections_search_builder_class.new(self, nil).tap do |builder|
```
jcoyne: remove this stuff if it's no longer needed.

app/search_builders/curation_concerns/collection_search_builder.rb
```diff
@@ -8,27 +8,31 @@ module CurationConcerns
...
     # Defines which search_params_logic should be used when searching for Collections
-    def initialize(context)
+    # @param [#blacklight_config, #current_ability] scope the object that has access to #blacklight_config
+    #   From a controller, scope would be `self`.  This argument is passed to ::SearchBuilder.new()
+    # @param [Symbol] permissions defining breadth of search, e.g. :edit, :read
+    # @note permissions will otherwise be defaulted by inherited #discovery_permissions
+    def initialize(scope, access = nil)
       @rows = 100
-      super(context)
+      @discovery_permissions ||= access_levels[access] if access
```
jcoyne: `@discovery_permissions` is always `nil` here, so why use `||=`?

atz: Not necessarily.  If this class is extended and *that* class' initializer sets the @discovery_permissions before calling this one, then we keep those (e.g. bypassing `access_levels`).  But you are right we can probably do without it.  

You can mostly ignore this PR while it is still WIP.

PR #1112, opened 2016-11-29
Removing length limit on identifier PairTree path

escowles:
Removes the current limit of 4 PairTree directory levels when storing uploaded files.  This prevents collisions when uploading files and using identifiers that are more than 8 characters long.

### DIFF
--- a/app/services/curation_concerns/working_directory.rb
+++ b/app/services/curation_concerns/working_directory.rb
@@ -48,7 +48,7 @@ module CurationConcerns
         end
 
         def full_filename(id, original_name)
-          pair = id.scan(/..?/).first(4)
+          pair = id.scan(/..?/)
           File.join(CurationConcerns.config.working_path, *pair, original_name)
         end
     end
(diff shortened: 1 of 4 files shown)


app/services/curation_concerns/working_directory.rb
```diff
@@ -48,7 +48,7 @@ def copy_stream_to_working_directory(id, name, stream)
         end
 
         def full_filename(id, original_name)
-          pair = id.scan(/..?/).first(4)
+          pair = id.scan(/..?/)
```
cbeer: This has the potential to create very deep directory structures, doesn't it? To avoid this, might it be better to add the original id as the last leaf node?

mjgiarlo: @escowles I rather like that approach. Mind putting in a follow-up PR to amend?

escowles: Yep, I'll do that now.

PR #1119, opened 2016-12-12
Remove dead code

jcoyne:
This is a private method that is not referenced.


jeremyf:
In digging through the code I see a usage of `curation_concern.in_collection_ids`; Which I believe is coming from the relationship `:in_collection` (and is clearly not a private method call).

https://github.com/projecthydra/curation_concerns/blob/master/app/actors/curation_concerns/actors/actor_stack.rb#L31

PR #1126, opened 2017-01-04
Adding check for presenter since it is required for the view

cam156:
refs projecthydra/sufia/issues/2990

If we are displaying the unavailable view and the workflow is in a suppressed state then the presenter does not get set in the usual way becuase document_not_found! raise an exception during the setting or the presenter.  At this point we need the presenter to be set for the view to render correctly, so set it without checking for suppression.

### DIFF
--- a/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
+++ b/app/controllers/concerns/curation_concerns/curation_concern_controller.rb
@@ -218,6 +218,7 @@ module CurationConcerns
         message = I18n.t("curation_concerns.workflow.unauthorized")
         respond_to do |wants|
           wants.html do
+            unavailable_presenter
             flash[:notice] = message
             render 'unavailable', status: :unauthorized
           end
@@ -235,6 +236,10 @@ module CurationConcerns
             render plain: message, status: :unauthorized
           end
         end
+      end
+
+      def unavailable_presenter
+        @presenter ||= show_presenter.new(::SolrDocument.find(params[:id]), current_ability, request)
       end
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/curation_concerns/curation_concern_controller.rb
```diff
@@ -218,6 +218,9 @@ def render_unavailable
         message = I18n.t("curation_concerns.workflow.unauthorized")
         respond_to do |wants|
           wants.html do
+            if @presenter.blank?
```
mjgiarlo: @jcoyne @cam156 How would you feel about:

1. Putting line 222 behind its own private message, one named something like `presenter_for_not_found_concern`; and
2. Invoking the new method within the `document_not_found!` method, which will make sure `@presenter` gets set right next to the code that handles the "not found" case

:question:

jcoyne: @mjgiarlo this partial was your creation, so I defer to you.

jcoyne: That said, it makes sense to have a really minimal presenter that only has the workflow state and the object type and label.

jcoyne: @cam156 that's right. So we need a SuppressedPresenter. The ShowPresenter is inappropriate as we're not showing the full representation

cam156:
@jcoyne & @mjgiarlo moved the code into a method.  Moving it down into an ensure in presenter is also an option, but I think that would seem more mysterious.

jcoyne:
@cam156 can you rebase this please?

PR #1134, opened 2017-01-13
Stop add_to_collection from removing works from other users' collecti…

hackmastera:
…ons fixes projecthydra/sufia#3011

port forward b812826280988be98196e473047e72d55813003e

### DIFF
--- a/app/actors/curation_concerns/actors/add_to_collection_actor.rb
+++ b/app/actors/curation_concerns/actors/add_to_collection_actor.rb
@@ -22,8 +22,11 @@ module CurationConcerns
           # remove from old collections
           (curation_concern.in_collection_ids - new_collection_ids).each do |old_id|
             collection = ::Collection.find(old_id)
-            collection.members.delete(curation_concern)
-            collection.save
+            # don't remove from collections current user doesn't have edit permissions for
+            if collection.edit_users.include? user.email
+              collection.members.delete(curation_concern)
+              collection.save
+            end
           end
 
           # add to new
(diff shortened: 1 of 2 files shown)


app/actors/curation_concerns/actors/add_to_collection_actor.rb
```diff
@@ -22,8 +22,11 @@ def add_to_collections(new_collection_ids)
           # remove from old collections
           (curation_concern.in_collection_ids - new_collection_ids).each do |old_id|
             collection = ::Collection.find(old_id)
-            collection.members.delete(curation_concern)
-            collection.save
+            # don't remove from collections current user doesn't have edit permissions for
+            if collection.edit_users.include? user.email
```
atz: Same question as https://github.com/projecthydra/curation_concerns/pull/1132/files#r95906550

Wouldn't you need a way for admins to delete also?

hackmastera: That seems like it would be useful. Are admins not automatically able to edit any collection? I'm not sure exactly how that gets set up.

hackmastera: my instinct is to say that is outside the scope of this PR. Before this PR, works were being deleted from collections as a side-effect of another user saving the work. This fixes that behavior. If certain users need fine-grained deletion abilities, I propose that should be dealt with at another time.

atz: OK, that seems reasonable.

jcoyne: Instead of using `email` use `user_key`, becuase some institutions (PSU) use `username` as the surrogate key instead of email. the `user_key` method encapsulates this variablity.

mjgiarlo: Was this PR merged before the email/user_key change was made? @hackmastera would you mind putting in a quick follow-up PR?

PR #1135, opened 2017-01-13
Stop add_as_member_of_collections_actor from removing works from othe…

hackmastera:
…r users' collections fixes projecthydra/sufia#3011

Also adds back some tests for this actor.

### DIFF
--- a/app/actors/curation_concerns/actors/add_as_member_of_collections_actor.rb
+++ b/app/actors/curation_concerns/actors/add_as_member_of_collections_actor.rb
@@ -16,7 +16,10 @@ module CurationConcerns
         # Maps from collection ids to collection objects
         def add_to_collections(collection_ids)
           return true unless collection_ids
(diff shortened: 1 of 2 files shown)


app/actors/curation_concerns/actors/add_as_member_of_collections_actor.rb
```diff
@@ -16,7 +16,10 @@ def update(attributes)
         # Maps from collection ids to collection objects
         def add_to_collections(collection_ids)
           return true unless collection_ids
+          # grab/save collections this user has no edit access to
+          other_collections = curation_concern.member_of_collections.select { |coll| !coll.edit_users.include?(user.email) }
```
jcoyne: instead of `email` can you use `user_key`? Otherwise this will break institutions like PSU who use `username` as the primary key.

mjgiarlo:
@jcoyne re-review?

PR #1137, opened 2017-01-17
Fix work presenter's solr query to get all file set ids, fixes #1136

### LINKED ISSUE #1136: SystemStackError in work_show_presenter
looks like this: 
```
SystemStackError (stack level too deep):
  app/views/curation_concerns/base/show.html.erb:16:in `_app_views_curation_concerns_base_show_html_erb___3900729328614262446_70171586185720'
```

Turns out solr is only returning 10 file sets so if your representative_id is not one of  those 10 the presenter thinks it's not a file set and keeps trying to drill down to get the file set.

patch forthcoming

hackmastera:

### DIFF
--- a/spec/presenters/curation_concerns/work_show_presenter_spec.rb
+++ b/spec/presenters/curation_concerns/work_show_presenter_spec.rb
@@ -81,6 +81,16 @@ describe CurationConcerns::WorkShowPresenter do
       expect(presenter.file_set_presenters.map(&:id)).to eq obj.ordered_member_ids
     end
 
+    context "solr query" do
+      before do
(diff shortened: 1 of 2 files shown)


spec/presenters/curation_concerns/work_show_presenter_spec.rb
```diff
@@ -93,6 +93,20 @@
       end
     end
 
+    context "when there are >10 file sets" do
+      before do
+        12.times do
+          fs = create(:file_set)
```
hackmastera: Sure, that would reduce the scope of the test and assume the indexing piece is tested sufficiently elsewhere.

PR #1141, opened 2017-02-01
Show a default image for files when user doesn't have read access

hortongn:
Fixes https://github.com/projecthydra/sufia/issues/2394

Shows a default.png instead of broken images when a user has read access to a work, but not the file(s).


app/controllers/concerns/curation_concerns/download_behavior.rb
```diff
@@ -40,6 +40,12 @@ def derivative_download_options
       # that files are in a LDP basic container, and thus, included in the asset's uri.
       def authorize_download!
         authorize! :read, params[asset_param_key]
+      rescue CanCan::AccessDenied
+        redirect_to default_image
+      end
+
+      def default_image
+        ActionController::Base.helpers.image_path 'default.png'
```
jcoyne: As this is a controller, can you just use `image_path 'default.png'`?

hortongn: @jcoyne I initially tried that but got `*** NoMethodError Exception: undefined method 'image_path' for #<DownloadsController:0x0000000a898ac8>`

PR #1149, opened 2017-02-09
Better error message for an etag conflict

jcoyne:


hackmastera:
@jcoyne Oh I think we need one last period at the end.

PR #1153, opened 2017-02-21
Cast proxy_for_id to integer before joining on an integer field

### LINKED ISSUE #1152: Type mismatch in PermissionQuery
When I try to transition an object in a Sipity workflow, I get an error in PermissionQuery.scope_users_for_entity_and_roles: "ActiveRecord::StatementInvalid (PG::UndefinedFunction: ERROR:  operator does not exist: integer = character varying", because there is a mismatch between the user.id column (integer) and the subquery (which produces varchars).  Note: this error doesn't happen with SQLite, but does happen with PostgreSQL.

The query is being executed here: https://github.com/projecthydra/curation_concerns/blob/master/app/services/curation_concerns/workflow/permission_query.rb#L267

We should be able to explicitly cast the user.id value to a varchar.

escowles:
Fixes #1152 

Porting https://github.com/projecthydra-labs/hyrax/commit/094c540124cb501f35a71e312954acbbdb13179c

### DIFF
--- a/app/services/curation_concerns/workflow/permission_query.rb
+++ b/app/services/curation_concerns/workflow/permission_query.rb
@@ -257,7 +257,10 @@ module CurationConcerns
             .and(entity_responsibilities[:entity_id].eq(entity.id))
         )
 
-        sub_query_for_user = agent_table.project(agent_table[:proxy_for_id]).where(
+        # PostgreSQL requires an explicit cast from string to integer
+        cast = Arel::Nodes::NamedFunction.new "CAST", [agent_table[:proxy_for_id].as("integer")]
+
+        sub_query_for_user = agent_table.project(cast).where(
           agent_table[:id].in(workflow_agent_id_subquery)
             .or(agent_table[:id].in(entity_agent_id_subquery))
         ).where(


app/services/curation_concerns/workflow/permission_query.rb
```diff
@@ -257,7 +257,10 @@ def scope_users_for_entity_and_roles(entity:, roles:)
             .and(entity_responsibilities[:entity_id].eq(entity.id))
         )
 
-        sub_query_for_user = agent_table.project(agent_table[:proxy_for_id]).where(
+        # PostgreSQL requires an explicit cast from string to integer
+        cast = Arel::Nodes::NamedFunction.new "CAST", [agent_table[:proxy_for_id].as("integer")]
```
cbeer: I don't think I understand what this query does, but this strikes me as a very low-level solution. Is there anything we can do at a higher level?

escowles: I believe the core issue here is that Sipity agent ids are strings, to allow for some magic ids such as "admin" or "registered".  But this query should only retrieve the integer ids that are linked to roles and works.  (If that's not true, then casting the user.id to string would be better).

I don't know if @jcoyne has more background on this from the original fix in Hyku.

cbeer: Ah, sorry, I didn't see that this was simply backporting a bug fix, and we certainly should just ship it. I don't know where we  ought to move this conversation, though.

escowles: I've created a Hyrax issue for this: https://github.com/projecthydra-labs/hyrax/issues/453

cbeer: 👍

PR #1154, opened 2017-02-23
prevents child relationship when parent is owned by a different depositor

revgum:
First step in surfacing a fix for Sufia Issue 3042: https://github.com/projecthydra/sufia/issues/3042

### DIFF
--- a/app/actors/curation_concerns/actors/add_to_work_actor.rb
+++ b/app/actors/curation_concerns/actors/add_to_work_actor.rb
@@ -22,13 +22,17 @@ module CurationConcerns
             work.save
           end
 
-          # add to new
+          # add to new so long as the depositor for the parent and child matches, otherwise inject an error
           (new_work_ids - curation_concern.in_works_ids).each do |work_id|
             work = ::ActiveFedora::Base.find(work_id)
-            work.ordered_members << curation_concern
-            work.save
+            if work.depositor != curation_concern.depositor
+              curation_concern.errors[:in_works_ids] << "Works can only be related to each other if they were deposited by the same user."
+            else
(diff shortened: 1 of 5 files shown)


revgum:
@vantuyls @straleyb and I worked a bit and explored how to try and address this using a validator, but they fired before the actors ran. The `AddToWorkActor` is responsible for persisting the relationship from parent to child.. and it was found that we can inject errors on the `curation_concern` to propagate them to the front-end. So this PR seems like the cleanest approach to getting validation as a byproduct of this persist.

It wasn't clear how else to accomplish this functionality without what seems like significantly reworking what and how the actor operates in conjunction with validations on the backend.

app/actors/curation_concerns/actors/add_to_work_actor.rb
```diff
@@ -22,13 +22,17 @@ def add_to_works(new_work_ids)
             work.save
           end
 
-          # add to new
+          # add to new so long as the depositor for the parent and child matches, otherwise inject an error
           (new_work_ids - curation_concern.in_works_ids).each do |work_id|
             work = ::ActiveFedora::Base.find(work_id)
-            work.ordered_members << curation_concern
-            work.save
+            if work.depositor != curation_concern.depositor
```
mjgiarlo: This line may come back to bite us, especially for folks with use cases where all depositors (in effect) share ownership of all objects. Thinking here of how Princeton and IU use CC. Ping @tpendragon @escowles 

I wonder if we should do an ability check here (delegating this responsibility to our Ability models and underlying hydra-access-controls code that knows how to inspect `edit_users` and `edit_groups`) instead of comparing depositors.

PR #1156, opened 2017-02-27
prevents user from adding ordered_member_ids which they do not own

revgum:
Adding a child work which is not owned by the current user fix for Sufia Issue 3042: projecthydra/sufia#3042

### DIFF
--- a/app/actors/curation_concerns/actors/apply_order_actor.rb
+++ b/app/actors/curation_concerns/actors/apply_order_actor.rb
@@ -3,8 +3,7 @@ module CurationConcerns
     class ApplyOrderActor < AbstractActor
       def update(attributes)
         ordered_member_ids = attributes.delete(:ordered_member_ids)
-        sync_members(ordered_member_ids)
-        apply_order(ordered_member_ids) && next_actor.update(attributes)
+        sync_members(ordered_member_ids) && apply_order(ordered_member_ids) && next_actor.update(attributes)
       end
 
       private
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@revgum There's not a very strong "ownership" model at the moment. Should the users listed in the object's `edit_users` list, and users in the `edit_groups` list, also be able to add works to works?

revgum:
@mjgiarlo This is a great question, and somewhat touches on a discussion that @vantuyls and I were having this morning. ApplyOrderActor and AddToWorkActor both cover similar functionality with regard to associating a work to another work (in the parent, or child context) and both would need this type of ownership consideration you're mentioning. 

If we can come to a consensus, should the same `edit_users` and `edit_groups` check happen for both actors?

mjgiarlo:
What do you think, @jcoyne?

app/actors/curation_concerns/actors/apply_order_actor.rb
```diff
@@ -20,10 +19,14 @@ def sync_members(ordered_member_ids)
 
           (ordered_member_ids - existing_members_ids).each do |work_id|
             work = ::ActiveFedora::Base.find(work_id)
-            curation_concern.ordered_members << work
+            if work.depositor != curation_concern.depositor
+              curation_concern.errors[:ordered_member_ids] << "Works can only be related to each other if they were deposited by the same user."
```
tpendragon: This seems like a very strict use case specific to maybe one or two institutions. Shouldn't this kind of logic sit in `ability.rb`? For instance, Plum has no concept of "the person uploading the file" as being of any importance whatsoever.

mjgiarlo: @tpendragon :+1: See my comment here: https://github.com/projecthydra/curation_concerns/pull/1156#issuecomment-282814247

revgum: my mistake @jcoyne.
Maybe this functionality really belongs directly in Sufia, @tpendragon @jcoyne?

jcoyne: Doesn't matter where it goes, we still need to encapsulate what users can do in the Ability class. That gives downstream users a single place to change what a user can/cannot do.

PR #1173, opened 2017-03-29
Allow full redis urls

### LINKED ISSUE #1172: Allow full redis urls
### Descriptive summary
The current implementation of the redis.yml parser does not support a full redis url. This prevents the usage of a password and alternate redis dbs.

### Expected behavior
allow `config/redis.yml` with the format:
```
default: &default
  url: "redis://:password@redis.example.com:6379/2"
```
### Actual behavior
`config/redis.yml` currently requires the following:
```
default: &default
  scheme: redis
  host: redis.example.com
  port: 6379
```

PR incoming...

nestorw:
Fixes #1172 

Allow a full url for redis connections.

This adds support for a new `url` option in `config/redis.yml` and won't break existing configurations.
The current parser doesn't allow for password and db number.

### DIFF
--- a/app/services/curation_concerns/lock_manager.rb
+++ b/app/services/curation_concerns/lock_manager.rb
@@ -27,8 +27,7 @@ module CurationConcerns
 
       def uri
         @uri ||= begin
-          opts = options
-          opts[:url].blank? ? URI("#{opts[:scheme]}://#{opts[:host]}:#{opts[:port]}").to_s : URI(opts[:url]).to_s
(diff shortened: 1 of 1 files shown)


app/services/curation_concerns/lock_manager.rb
```diff
@@ -28,7 +28,7 @@ def lock(key)
       def uri
         @uri ||= begin
           opts = options
```
jeremyf: I don't believe this `opts = options` is getting us anything (except an additional local variable assignment). Do you have time to tweak this?

nestorw:
Good point, `opts` removed.

PR #1181, opened 2017-04-24
Bump to version 1.7.7

hortongn:


lib/curation_concerns/version.rb
```diff
@@ -1,3 +1,3 @@
 module CurationConcerns
-  VERSION = '1.7.5'.freeze
+  VERSION = '1.7.7'.freeze
```
jrochkind: What happened to 1.7.6?

hortongn: 1.7.6 got realeased without the version change getting merged back into the 1-7-stable branch for some reason.  The current version is definitely 1.7.6.  So bumping to 1.7.7 is correct.

jrochkind:
@hortongn I'm dismayed that I seemed to get an email from travis saying this did not pass CI. But I'm super confused what's going on. Any thoughts?

hortongn:
@jrochkind curation_concerns has at least one test that tends to be flaky on Travis.  Sometimes we need to re-run Travis one or two times to get all the tests green.  So if you ever see one or two specs failing, just restart the Travis build.

hackmastera:
The 1.7 branch appears to have the same testing problem that affected the 2.0 branch for a long time; if we want to resolve that failing spec we could backport this commit https://github.com/projecthydra/curation_concerns/commit/e3b36d79b0b28d484f17405160423de17ffaad95

PR #1183, opened 2017-07-20
Reverted i18n gem to version 0.8.1

smithjp:
Fixes test failures due to missing translations.

It appears that the curation_concerns gem is not compatible with i18n gem >= 0.8.3, so an explicit dependency for i18n version 0.8.1 is being added.

@samvera/sufia-code-reviewers


awead:
Duplicate of #1184

PR #1184, opened 2017-07-20
Reverted i18n gem to version 0.8.1

smithjp:
Fixes test failures due to missing translations.

It appears that the curation_concerns gem is not compatible with i18n gem >= 0.8.3, so an explicit dependency for i18n version 0.8.1 is being added.

@samvera/sufia-code-reviewers

### DIFF
--- a/spec/test_app_templates/Gemfile.extra
+++ b/spec/test_app_templates/Gemfile.extra
@@ -8,7 +8,7 @@ group :development do
   gem 'better_errors'
   gem 'binding_of_caller'
 
-  # This change line can be removed once issue at 
+  # This line can be removed once issue at 
   # https://github.com/svenfuchs/i18n/issues/379 is resolved
   gem 'i18n', '0.8.1'
 end
--- a/spec/test_app_templates/Gemfile.extra
+++ b/spec/test_app_templates/Gemfile.extra
@@ -7,5 +7,8 @@ end
 group :development do
   gem 'better_errors'
   gem 'binding_of_caller'
+
+  # This change line can be removed once issue at 
+  # https://github.com/svenfuchs/i18n/issues/379 is resolved
   gem 'i18n', '0.8.1'
 end
--- a/spec/test_app_templates/Gemfile.extra
+++ b/spec/test_app_templates/Gemfile.extra
@@ -7,4 +7,5 @@ end
 group :development do
   gem 'better_errors'
   gem 'binding_of_caller'
+  gem 'i18n', '0.8.1'
 end


jcoyne:
Can you provide a replication case for the bug? Have you filed an issue with the maintainers of the i18n gem?

jcoyne:
Was it this issue? https://github.com/CanCanCommunity/cancancan/issues/415

smithjp:
The test failures look like this:

     ActionView::Template::Error:
       translation missing: en.false
     # ./app/renderers/curation_concerns/renderers/configured_microdata.rb:13:in `microdata_object?'
     # ./app/renderers/curation_concerns/renderers/configured_microdata.rb:17:in `microdata_object_attributes'
     # ./app/renderers/curation_concerns/renderers/attribute_renderer.rb:28:in `render'
     # ./app/presenters/curation_concerns/presents_attributes.rb:20:in `attribute_to_html'
     # ./app/views/curation_concerns/base/_attributes.html.erb:8:in `__home_travis_build_samvera_curation_concerns_app_views_curation_concerns_base__attributes_html_erb__2674987801978955349_140047760'
     # ./app/views/curation_concerns/base/show.html.erb:19:in `__home_travis_build_samvera_curation_concerns_app_views_curation_concerns_base_show_html_erb___168103936029684779_140777240'
     # ./spec/features/add_file_spec.rb:18:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # I18n::MissingTranslationData:
     #   translation missing: en.curation_concerns.schema_org.permission_badge.type
     #   ./app/renderers/curation_concerns/renderers/configured_microdata.rb:13:in `microdata_object?'

Which can be seen in the automated tests for previous pull request at: https://travis-ci.org/samvera/curation_concerns/jobs/240798479.

jcoyne:
@smithjp to me it looks like the problem is on this line: https://github.com/samvera/curation_concerns/blob/master/app/renderers/curation_concerns/renderers/configured_microdata.rb#L13

The default probably shouldn't be `false`.  Maybe empty string or nil?

jcoyne:
I think I see what is happening here:

When you say `t('foo', default: false)` the TranslationHelper in rails actually calls:
`I18n.translate('foo', default: [false], raise: true)`.  I think this was broken by this change:
https://github.com/svenfuchs/i18n/commit/1641e5fefebf53435bcc45eacf1e4659d46dece9

I made an upstream issue: https://github.com/svenfuchs/i18n/issues/379

Please add a comment next to your change that points to that issue. That way we can back out your change when the issue is resolved upstream.

smithjp:
I have added a comment in the Gemfile.extra file.  Is this ok?

PR #1186, opened 2017-09-12
Allow recent kaminari_route_prefix (1-7-stable)

jrochkind:
Which will then allow kaminari 1.0.

Not sure what the requirement should actually be, so I just said <2, figuring
eventually kaminari_route_prefix 1.0 is likely to be backwards compat with
current (already in production in many apps) 0.1.1

PR'd to 1-7-stable for 1.7.8 release with this relaxation.

### DIFF
--- a/curation_concerns.gemspec
+++ b/curation_concerns.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'hydra-editor', '>= 2', '< 4'
   spec.add_dependency 'rails_autolink'
   spec.add_dependency 'sprockets-es6'
-  spec.add_dependency 'kaminari_route_prefix', '~> 0.0.1'
+  spec.add_dependency 'kaminari_route_prefix', "< 2"
   spec.add_dependency 'active_attr'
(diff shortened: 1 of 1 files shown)


jrochkind:
do not understand why these are failing, seems engine-cart related and unrelated to the kaminari_route_prefix bump. 

I do not have time to debug engine cart right now, so it goes.

jcoyne:
Looks like the failures are due to using a version of Blacklight that generates Rails 5.1 compatible migrations (which means not Rails 4.2 compatible)

jrochkind:
@jcoyne, do you think it's okay to merge and release a 1.7.8 even with failing tests due to that issue?

jrochkind:
@mjgiarlo , can you give me some agreement that it seems okay to merge even with failing tests that were already failing? I want someoen to say they think it's okay too. :)

mjgiarlo:
@jrochkind fine by me

jrochkind:
Actually github will not let me merge, I do not have the privs when travis is failing. @mjgiarlo, can you as a favor?

